Issue #32, opened 2017-04-26
Add retina support to raster tiles (@2x) + texture compression in Unity


david-rhodes:
Our API supports [requesting textures at 2x](https://www.mapbox.com/api-documentation/#retrieve-a-static-map-from-a-style), but our SDK does not help with formatting the URL to do so. 

We should also offer the ability to [compress at runtime](https://docs.unity3d.com/ScriptReference/Texture2D.Compress.html) in Unity to save memory (at the cost of CPU time during compression), and investigate the [ability to load raw data in a compressed format](https://docs.unity3d.com/ScriptReference/Texture2D.LoadRawTextureData.html).

david-rhodes:
Done: https://github.com/mapbox/mapbox-unity-sdk/pull/48, https://github.com/mapbox/mapbox-sdk-cs/pull/77

david-rhodes:
_Original_
<img width="536" alt="screen shot 2017-05-02 at 10 13 55 am" src="https://cloud.githubusercontent.com/assets/23202691/25684825/bb2b5c06-3020-11e7-9eb0-38d4b49fd217.png">

_Retina_
<img width="542" alt="screen shot 2017-05-02 at 10 13 29 am" src="https://cloud.githubusercontent.com/assets/23202691/25684823/b8fdb1f4-3020-11e7-83d9-e8a0aa0c257f.png">

david-rhodes:
Reopening because I just found out that the final texture format changes automatically after `texture.LoadImage();`. In order to actually support different types, we may need to investigate `LoadRawTextureData`. 

Worst case scenario, I'll remove texture format and just have a boolean for `compress textures`.

PR #48, opened 2017-05-03
Retina raster

david-rhodes:
Updating submodule for retina tiles
Changed editor and factory to support new texture settings.


david-rhodes:
See: https://github.com/mapbox/mapbox-unity-sdk/issues/32

PR #53, opened 2017-05-05
Cutting holes

brnkhy:
@david-rhodes can you check this if/when you have time please?
This'll probably be a little hard to merge in as it's a quite big PR but would be nice if we can test&talk about it next week.


david-rhodes:
- Not all examples updated to use new modules (`Drive` and `Slippy` are broken)
- `Snap y to zero` (`MapController`) does not seem to be working
    - Use `Boulder, Co`, for `MeshGeneration.unity` location and you will see the tiles are too far above the ground plane to see in camera view
- Modifiers need to have a static order (custom inspectors can even reorder as they are added)
    - We could have always benefitted from this, but now that we have even more modifiers, they need to be ordered properly (by us).
- Roofs on terrain are still not straight:
screen shot 2017-05-08 at 2 12 03 pm

On the up side, there are zero conflicts when trying to pull develop in 👍.

brnkhy:
@david-rhodes I fixed a small issue with y axis coordinates.
flat roofs should be OK now, you sure you had `Flat Roof` checked in height modifier?
unity_2017-05-15_17-44-52

demos might be totally broken, files are in bad/wrong folders as well. I'll polish this and get ready to merge asap.
`snap y` is globally broken I guess, I'll check that as well.

brnkhy:
@david-rhodes I fixed some scenes and module references. should be easier to review now.
I had different plans for modifier hierarchy/order but looks like it'll actually be much better if we switch to a much more restrictive system for that in short term. I have some ideas for that, let's talk whenever you have time.

PR #57, opened 2017-05-09
add "Add Collider" and "Add To Layer" properties to terrain factory

brnkhy:
add "Add Collider" and "Add To Layer" properties to terrain factory


sdkproject/Assets/Mapbox/Core/Unity/MeshGeneration/Factories/TerrainFactory.cs
```diff
@@ -250,9 +258,10 @@ private void CreateFlatMesh(UnityTile tile)
             if (tile.MeshRenderer.material == null)
                 tile.MeshRenderer.material = _baseMaterial;
 
-            //BRNKHY Optional stuff
-            //go.AddComponent<MeshCollider>();
-            //go.layer = LayerMask.NameToLayer("terrain");
+            if (_addMeshCollider)
+                tile.gameObject.AddComponent<MeshCollider>();
```
david-rhodes: If this is a flat mesh, can we instead use a box collider @brnkhy?

sdkproject/Assets/Mapbox/Core/Unity/MeshGeneration/Factories/TerrainFactory.cs
```diff
@@ -43,6 +43,12 @@ public class TerrainFactory : Factory
         private float _heightModifier = 1f;
         [SerializeField]
         private int _sampleCount = 40;
+        [SerializeField]
+        private bool _addMeshCollider = false;
+        [SerializeField]
+        private bool _addToLayer = false;
+        [SerializeField]
+        private int _layerId = 0;
```
david-rhodes: Should the layer option be in the base class? It makes sense for all factories to have the ability to specify which layer they place objects in. Further down the chain, specific stacks can offer layer overrides?

sdkproject/Assets/Mapbox/Core/Unity/MeshGeneration/Factories/TerrainFactory.cs
```diff
@@ -214,9 +220,11 @@ private void GenerateTerrainMesh(UnityTile tile, float heightMultiplier)
 
             if (tile.MeshRenderer.material == null)
                 tile.MeshRenderer.material = _baseMaterial;
-            //BRNKHY Optional stuff
-            //go.AddComponent<MeshCollider>();
-            //go.layer = LayerMask.NameToLayer("terrain");
+
+            if (_addMeshCollider)
```
david-rhodes: Pedantic, but please always use braces.

sdkproject/Assets/Mapbox/Core/Unity/Editor/TerrainFactoryEditor.cs
```diff
@@ -27,8 +30,10 @@ void OnEnable()
         heightMod_Prop = serializedObject.FindProperty("_heightModifier");
         mapId_Prop = serializedObject.FindProperty("_mapId");
         material_Prop = serializedObject.FindProperty("_baseMaterial");
-
-        customMapId_Prop = serializedObject.FindProperty("_customMapId");
+        collider_Prop = serializedObject.FindProperty("_addMeshCollider");
+        addLayer_Prop = serializedObject.FindProperty("_addToLayer");
+        addLayer_Prop = serializedObject.FindProperty("_addToLayer");
```
david-rhodes: duplicate line?

sdkproject/Assets/Mapbox/Core/Unity/MeshGeneration/Factories/TerrainFactory.cs
brnkhy: not all factories actually create game objects (like map image factory). And in mesh factory, it would make much more sense to have layer visualizers have this instead of the factory OR even better, have a layer modifier etc. 
so I'm not sure about moving this down to base class

PR #63, opened 2017-05-11
Prevent super tall pillars from appearing on oceans where there is no elevation data

Spaxe:
If @david-rhodes could review this, that would be awesome.

### DIFF
--- a/sdkproject/Assets/Mapbox/Core/Unity/MeshGeneration/Factories/TerrainFactory.cs
+++ b/sdkproject/Assets/Mapbox/Core/Unity/MeshGeneration/Factories/TerrainFactory.cs
@@ -153,22 +153,19 @@ namespace Mapbox.Unity.MeshGeneration.Factories
                     var xx = Mathd.Lerp(tile.Rect.Min.x, tile.Rect.Max.x, xrat);
                     var yy = Mathd.Lerp(tile.Rect.Min.y, tile.Rect.Max.y, yrat);
 
-                    var height = heightMultiplier * Conversions.GetRelativeHeightFromColor(tile.HeightData.GetPixel(
-                        (int)(xrat * 255),
-                        (int)((1 - yrat) * 255)),
-                        tile.RelativeScale);
-
-                    /*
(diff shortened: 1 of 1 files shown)


sdkproject/Assets/Mapbox/Core/Unity/MeshGeneration/Factories/TerrainFactory.cs
```diff
@@ -153,12 +153,22 @@ private void GenerateTerrainMesh(UnityTile tile, float heightMultiplier)
...
+                    var height = heightMultiplier * Conversions.GetRelativeHeightFromColor(tile.HeightData.GetPixel(
+                        (int)(xrat * 255),
+                        (int)((1 - yrat) * 255)),
+                        tile.RelativeScale);
+
+                    /*
+                        Assuming this is in meters? The tallest point on earth is ~8,500m.
+                        Therefore this number should be pretty safe.
+                    */
+                    if (height > 100000 * tile.RelativeScale * heightMultiplier) {
```
david-rhodes: This will likely work. However, we have determined the missing data is because of bad rendering on our side. Textures with missing data will have width less than 256 (usually 8?). Therefore, it is likely more correct to set `heightMultiplier` to zero in `CreateTerrainHeight()`:

```cs
var texture = new Texture2D(256, 256);
texture.wrapMode = TextureWrapMode.Clamp;
texture.LoadImage(pngRasterTile.Data);
if (texture.width < 256)
{
	heightMultiplier = 0f;
}
tile.HeightData = texture;
tile.HeightDataState = TilePropertyState.Loaded;
GenerateTerrainMesh(tile, heightMultiplier);
```

Spaxe:
Thanks @david-rhodes, this method seems to work.

Spaxe:
When I log into Travis CI, I can't see mapbox/mapbox-unity-sdk, so I don't know what build is failing or how.

(It works on my machine! ;)

david-rhodes:
@Spaxe Thanks for your patience. Travis is being strange, but I don't see anything in this PR that is risky. Merging.

PR #64, opened 2017-05-11
Coding style

david-rhodes:
First pass coding style. Link is set to develop for when this is merged.

Can we all agree on these? Should we make a note about fixing any existing code that does not already follow these standards?

### DIFF
--- a/CODING-STYLE.md
+++ b/CODING-STYLE.md
@@ -1,0 +1,40 @@
+- Use tabs, not spaces, for indentation.
+
+- Private fields should be prefixed with an underscore (`_privateField`) and generally avoid `this` keyword.
+
+- Use verbose variable names. Do not abbreviate.
+
+- Braces even for one liners:
+
+  **RIGHT**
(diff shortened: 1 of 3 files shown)


Spaxe:
As long as there are no mixed tabs with spaces in the same file, it should be OK?

BergWerkGIS:
> First pass coding style. Link is set to develop for when this is merged.

@david-rhodes I changed that to be relative and stay in the same branch.
I think it doesn't make sense to pin to `develop`.

> As long as there are no mixed tabs with spaces in the same file, it should be OK?

@Spaxe we'd like to have tabs in all code files.

david-rhodes:
@Spaxe the problem is that many of us use "auto-format" in our IDE. If we eventually modify a file that had pure spaces, the diff will show every line as being different. Makes changes hard to track.

binakot:
I hope this link helps for the issue http://www.dofactory.com/reference/csharp-coding-standards

PR #67, opened 2017-05-15
Update to latest Mapbox.VectorTile@v1.0.4-alpha2

BergWerkGIS:
update dependency `vector-tile-cs` to latest `Mapbox.VectorTile@v1.0.4-alpha2`
includes update to latest `mapbox-sdk-cs`.


brnkhy:
@BergWerkGIS tried running all demos with this, couldn't see any problems. I think this is all internal and shouldn't effect unity side much either right Willy?

BergWerkGIS:
@brnkhy yes, everything that reaches into `vector-tiles-cs` via `mapbox-sdk-cs` should be fine.

I only had to change `Assets/Mapbox/Core/Unity/MeshGeneration/Data/VectorFeatureUnity.cs` which uses `vector-tile-cs` directly.

PR #68, opened 2017-05-15
add 3d poi visualization layer for MeshGenerationPois demo

brnkhy:


david-rhodes:
@brnkhy Did you want to rename `ILabelVisualizationHelper` first, and maybe remove the need for `FeatureBehaviour`? It feels coupled, right now.

brnkhy:
@david-rhodes `feature behaviour` is different, I add that to everything and it just contains the raaw data for easier debugging. So it's not a poi thing.
`ILabelVisualizationHelper` is used to pass feature data to prefabs, that's all. And yes we should rename that, you like `IPoiPrefabHelper` ?

david-rhodes:
@brnkhy I just checked this and 3D poi is not default in the visualizer. Was it intended to be?

brnkhy:
@david-rhodes sorry I just realized I didn't include asset files in that commit, pushed them as well, it should be there now.

PR #84, opened 2017-05-23
Updating ignore to support meta and unity directory has been moved.

david-rhodes:
Android added to json meta and updated gitignore. 

I think we still have to remove submodule???


david-rhodes:
@BergWerkGIS is there a reason to use the dependencies folder JSON, or should we just include these files directly in `sdkproject`?

BergWerkGIS:
@david-rhodes not sure, personal preference would be to keep all "external" stuff separate but I'm ok with just keeping submodules in dependencies.

Issue #35, opened 2017-04-26
Tiles and maps built with the map factory framework should report progress


david-rhodes:
As a developer, I should know when a tile is finished loading (from the web), as well as any post-processing that happens after the tile is fetched (texture generated, meshes built, etc.). This will help to create animations, avoid temporal coupling, and allow for loading screens, etc.

david-rhodes:
First pass progress reporting has been implemented: #89.

The AbstractMap has an initialized event, but this is used for doing conversions that require specific map attributes (reference tile and world scale, etc.). The MapVisualizer now has state changed events, as well as all the TileFactories. Use these to know when specific factories, or the entire map are done loading.

PR #89, opened 2017-05-24
Progress report #35

brnkhy:
solves #35


david-rhodes:
@brnkhy Nevermind. I fixed it. Also fixed a few other bugs--want to check them out, please?

brnkhy:
@david-rhodes ah I think you're right, what I did would cause errors when flat terrain factory ends last. Tested a few demos as well, everything looks fine. Not sure if pure white terrain on `MeshGenerationOnTerrain` demo is intentional though, looks a bit ugly. other than that it'll all fine.

Issue #95, opened 2017-05-26
Enable configuration of tile caching systems


david-rhodes:
I imagine this will be part of the `Mapbox->Configure` menu. We should allow developers to choose which memory and file caching strategies they use, and how big they are.

david-rhodes:
This was done.

Issue #19, opened 2017-04-19
triage WebGL builds


isiyu:
WebGL builds of our examples aren't working in latest develop. 
@MateoV - do you have bandwidth to investigate this?

BergWerkGIS:
Looked at the `Slippy` example.

First I got: `InvalidTokenException: Please configure your access token in the menu!`
But `StreamingAssets` exists in the generated output.

I applied this patch:
```diff
diff --git a/sdkproject/Assets/Mapbox/Unity/MapboxAccess.cs b/sdkproject/Assets/Mapbox/Unity/MapboxAccess.cs
index ff6e2a7..611fe9a 100644
--- a/sdkproject/Assets/Mapbox/Unity/MapboxAccess.cs
+++ b/sdkproject/Assets/Mapbox/Unity/MapboxAccess.cs
@@ -93,7 +93,7 @@ namespace Mapbox.Unity

                private void ValidateMapboxAccessFile()
                {
-#if !UNITY_ANDROID
+#if !UNITY_ANDROID && !UNITY_WEBGL
                        if (!Directory.Exists(Application.streamingAssetsPath) || !File.Exists(_accessPath))
                        {
                                throw new InvalidTokenException("Please configure your access token in the menu!");
@@ -107,7 +107,7 @@ namespace Mapbox.Unity
                /// </summary>
                private void LoadAccessToken()
                {
-#if UNITY_EDITOR || !UNITY_ANDROID
+#if UNITY_EDITOR || (!UNITY_ANDROID && !UNITY_WEBGL)
                        AccessToken = File.ReadAllText(_accessPath);
 #else
             AccessToken = LoadMapboxAccess();
```

and the token seems to load fine now.

However the WebGL page has been loading for several minutes (+8) now.
Chrome's task manager shows (CPU constantly changing between ~26% and 33%):
![image](https://cloud.githubusercontent.com/assets/4549888/26576425/2f7a428a-4529-11e7-951c-c05a427fd805.png)
and I'm not even able to close the tab.

Will try a lighter example next.

BergWerkGIS:
`Playground/RasterTile` example:
* Same behavior as above.
* Tried a development build.
  * same behavior as above, except that the tab now uses almost **1.3GB** memory
  * last message I see in the console:
  ```
  warning: 2 FS.syncfs operations in flight at once, probably just doing extra work
  ```

BergWerkGIS:
[...]

PR #109, opened 2017-06-02
Fix for webgl and added additional configuration options

david-rhodes:
Using `Resources` instead of `StreamingAssets` to force synchronous file loading, which fixes WebGL issue.

Added memory cache size configuration option.

Test on WebGL, iOS, and Android--all appear to work great!


david-rhodes:
Reference: https://github.com/mapbox/mapbox-unity-sdk/issues/95, https://github.com/mapbox/mapbox-unity-sdk/issues/19

PR #117, opened 2017-06-08
ios-telemetry-v0.2.2

isiyu:
Embed bitcode, for cloudbuild


david-rhodes:
@isiyu I don't actually see any new files here. Am I missing something?

PR #119, opened 2017-06-09
Mesh factory fix

brnkhy:
there are bunch of fixes there but the biggest thing is mesh factory pre-downloading data I guess.


sdkproject/Assets/Mapbox/Unity/Map/RangeTileProvider.cs
```diff
@@ -5,15 +5,21 @@ namespace Mapbox.Unity.Map
...
+        private int _west;
+        [SerializeField]
+        private int _north;
+        [SerializeField]
+        private int _east;
+        [SerializeField]
+        private int _south;
 
-		internal override void OnInitialized()
+        internal override void OnInitialized()
```
BergWerkGIS: @brnkhy could you please use tabs instead of spaces?
Thanks.

PR #124, opened 2017-06-15
fix polygon modifier to support holes and multiple part features better

brnkhy:


david-rhodes:
Will be included in a future PR.

PR #125, opened 2017-06-16
MBTiles cache

BergWerkGIS:
This PR adds MBTiles caching to tile requests.

Had to change `FileSource.Request` to accept `map id` and ` tile id` instead of just the resource URL to create meaningful database names.

`CachingWebFileSource` does not use cache if `maxCacheSize == 0`.

Cache size setting is exposed in the Mapbox Configuration Window via Sliders.

I've tested on the following platforms so far:

* Unity Editor
* Windows native x64
* Windows UWP
* Android
* WebGL (no MBTiles caching there, just memory caching)


sdkproject/Assets/Mapbox/Core/mapbox-sdk-cs/Platform/Cache/CachingWebFileSource.cs
```diff
@@ -4,55 +4,131 @@
...
+		public IAsyncRequest Request(
+			string uri
+			, Action<Response> callback
+			, int timeout = 10
+			, CanonicalTileId tileId = new CanonicalTileId()
+			, string mapId = null
+		)
 		{
 
+			if (string.IsNullOrEmpty(mapId))
```
david-rhodes: @BergWerkGIS Do we need this here, or can we delegate to the cache itself? String checks aren't cheap, right?

BergWerkGIS: @david-rhodes what would be the benefit of delegating to the cache itself?

This check has to be done anyway and I'd prefer to fail as early as possible. 
I think it wouldn't make sense to fire a request and then throw after it has returned. Time and resources wasted.
Or am I missing something?

david-rhodes: I was just thinking we would be able to throw an exception without checking the string, but it's no big deal.

PR #129, opened 2017-06-21
Snap to origin

brnkhy:
Snap Map To Origin script, moving map in y axis and pulls it to zero. Should be used along side AbstractMap.


sdkproject/Assets/Mapbox/Unity/Map/AbstractMap.cs
```diff
@@ -126,7 +131,18 @@ protected virtual void Start()
 
 		void TileProvider_OnTileAdded(UnwrappedTileId tileId)
 		{
-			_mapVisualizer.LoadTile(tileId);
+			var utile = _mapVisualizer.LoadTile(tileId);
+			if (!_centerTileSet)
+			{
+				_centerTileSet = true;
+
+				// FIXME: this is never unregistered!
```
david-rhodes: This is never unregistered. If you try to move around when the tile is recycled, the map root transform will pop to the new height.

sdkproject/Assets/Mapbox/Unity/Map/SnapMapToOrigin.cs
```diff
@@ -0,0 +1,27 @@
...
+		{
+			Debug.Log("Snap Map to Origin script should be on the same object with AbstractMap");
+			return;
+		}
+
+		_map.OnHeightChanged += (s) =>
+		{
+			_map.Root.transform.position = new Vector3(
+				_map.Root.transform.position.x,
+				-s.CenterHeight * _map.Root.transform.localScale.x,
```
david-rhodes: Using local scale is "special knowledge" and should be avoided. Can't you use _map.WorldRelativeScale?
Also, won't this break terrain snapping modifier (the one that doesn't use raycasting)?

sdkproject/Assets/Mapbox/Unity/Map/AbstractMap.cs
```diff
@@ -126,7 +131,18 @@ protected virtual void Start()
 
 		void TileProvider_OnTileAdded(UnwrappedTileId tileId)
 		{
-			_mapVisualizer.LoadTile(tileId);
+			var utile = _mapVisualizer.LoadTile(tileId);
+			if (!_centerTileSet)
```
david-rhodes: Is there a better way to query the "map's center tile?"

brnkhy: I agree that's quite bad but couldn't think of anything better (without too much change) to detect the center of the map (first tile). I'll think about this again tomorrow though, it has been a while since I did this.

sdkproject/Assets/Mapbox/Unity/Map/SnapMapToOrigin.cs
brnkhy: @david-rhodes you're right, not sure why I did it like that there. I think I was thinking this more like  abstract map modifiers when I did this but even then, abstract map object doesn't have to be the root, so yes this is bad.

sdkproject/Assets/Mapbox/Unity/Map/AbstractMap.cs
brnkhy: This is about that "central tile detection" thing again. My main problem with this was, I wanted it to support central tile changes so you can shift the world in any direction. And if I remember correctly, I was thinking about creating a function to shift the world and move this to that function in the future. Then it won't be that utile object, but it'll be the one passed to that function. It might be better trying to implement that right away, even if we don't do the shifting thing though.

david-rhodes:
@brnkhy This design seems complicated and has several temporal coupling issues. For example, now that we have disk caching, the height data event is not registered before being provided. 

I feel like a better design would be to expose the tiles that the map currently knows about (via MapVisualizer), and let the developer query the height of any given tile (they can pick which one they want, in the event they want to "re-snap" a slippy map, or something).

PR #132, opened 2017-06-21
fix a bug where height is calculated wrong

brnkhy:


david-rhodes:
As discussed, this isn't a bug per se, but we should allow developers to use absolute (do not use `_relativeScale`) OR perceived height in their terrain factories.

PR #133, opened 2017-06-22
Port unit tests from sdk-cs

### LINKED ISSUE #96: Port unit tests from sdk-cs
Unity has a built-in editor test runner, but we can also add NUnit packages to our solutions to run tests there (just make sure tests are nested in an `Editor` folder).

We will also need to ensure our tests work with .NET 3.5.

BergWerkGIS:
closes #96 

Majority of tests carried over and ✅ (see below).

A few are disabled as I've implemented them as `Editor` tests and thus they end up in a different assembly than the `mapbox-sdk-cs` sources.
Being in different assemblies some classes cannot be accessed having the access modifier `internal`, eg [`TileResource`](https://github.com/mapbox/mapbox-unity-sdk/blob/unittests/sdkproject/Assets/Mapbox/Core/mapbox-sdk-cs/Map/TileResource.cs#L11)

Maybe we want to revisit the modifiers some time in the future.


BergWerkGIS:
image

PR #135, opened 2017-06-23
Refactor MapboxConfigurationWindow

### LINKED ISSUE #134: Refactor MapboxConfigurationWindow
Current implementation of `MapboxConfigurationWindow` hooks into `OnGui` to verify the token and write the configuration file.
This happens to often.
Refactor to reduce verification API calls and configuration file writing.

BergWerkGIS:
fixes #134 

**This is based off [PR `Port unit tests from sdk-cs`](https://github.com/mapbox/mapbox-unity-sdk/pull/133) which should be merged first.**

`MapboxConfigurationWindow` behaved a bit sluggish because of checking the token and saving the configuration file too often.

This PR checks the token only on load of config window or if it has changed and has the right length.

Likewise config file only gets saved when anything changes - including a delay of 500ms, meaning dragging the sliders works smoothly now.


david-rhodes:
@BergWerkGIS After loading the config window, my token is in the string field, but the validation string says that it is not valid (and it won't refresh).
screen shot 2017-06-23 at 9 38 01 am

Additionally, you can "save" an invalid token.

BergWerkGIS:
@david-rhodes I know you are busy with other things, but would be great if you could give this another spin some time this week:

* RegEx check for local pre-validation instead of just checking length
* doesn't save invalid token

isiyu:
@BergWerkGIS made 1 change to add `(ms)` to timeout field. 

I ran into this while testing the config window. The sliders are all responsive now, but found this bug when saving a valid token. Here, I had an invalid token saved, when I enter a valid token, it validates correctly and says it saves to the txt file, but closing the window and reopening, I'm getting the same old bad token. I assume it has something to do with the token not properly saving to the txt file in /streamingassets. 
https://www.dropbox.com/s/3wmfunq86viqbqj/unity_mapboxconfig.mov?dl=0

BergWerkGIS:
@isiyu changed your timeout change as [timeout is in seconds](https://github.com/mapbox/mapbox-unity-sdk/blob/2eaeff75e81de515c9d0c91ad879061e97604879/sdkproject/Assets/Mapbox/Unity/Utilities/HTTPRequest.cs#L62).

wrt to invalid token: 
I tried using an invalid token (changed config file manually then started Unity) about 15 times and wasn't able to reproduce the issue.
Nevertheless I added another safe measure that a new and valid token indeed does get saved.

-----

**But**, I've come across another issue:

If you open the project and immediately (_before Unity has finished loading and processing the project_) click on `Mapbox -> Configuration`:
* default values of [MapboxConfiguration](https://github.com/mapbox/mapbox-unity-sdk/blob/2eaeff75e81de515c9d0c91ad879061e97604879/sdkproject/Assets/Mapbox/Unity/MapboxAccess.cs#L200-L202) are displayed, not the ones from the config file
* if you enter a token (no matter if valid or invalid) the dialog is stuck at `Verifying token!`

I suppose that's some kind of race condition of different initializations and the use of `Runnable.Run(ValidateToken(_accessToken));` as the `_validating` flag that get's set to `true` at the beginning of `ValidateToken()` is always true afterwards, which means it never reaches `finally` where it would get reset to `false`.

@david-rhodes any ideas how to add the configuration menu only after the projects has fully loaded.
Or somehow determining if the project is still loading and only afterwards start the verification process.

david-rhodes:
@BergWerkGIS let me know if you think this will help solve the latest problem: https://docs.unity3d.com/ScriptReference/EditorApplication-isCompiling.html.

BergWerkGIS:
thanks @david-rhodes but `EditorApplication.isCompiling` but I couldn't get `EditorApplication.isCompiling` to work the way we need it.

Found [another solution](https://github.com/mapbox/mapbox-unity-sdk/pull/135/commits/a2900feff3230eb79b61cbc98bc3afc1c75da420) though and from my point of view we should be good to merge.

PR #144, opened 2017-07-06
Brnkhy sqlite meta fix

brnkhy:


BergWerkGIS:
> fix colliding sqlite dlls, it seems only one has to be marked as editor so converted x86 back

Does this work with 32bit Unity?
I think 32bit Unity might need the x86 dll.

image

brnkhy:
@BergWerkGIS hmm good question, probably won't work at 32bit version as you said. It just didn't let me change both to editor though, poped a few errors that I can't remember at the moment.

BergWerkGIS:
Check here too:
image

brnkhy:
@BergWerkGIS 
```Multiple plugins with the same name 'sqlite3' (found at 'Assets/Mapbox/Core/Plugins/sqlite/x64/sqlite3.dll' and 'Assets/Mapbox/Core/Plugins/sqlite/x86/sqlite3.dll'). That means one or more plugins are set to be compatible with Editor. Only one plugin at the time can be used by Editor.```
still getting this with two enabled for editor.

BergWerkGIS:
Works for me!

X86:
image

x64:
image

brnkhy:
@BergWerkGIS ah that makes totally sense

david-rhodes:
@BergWerkGIS I tried to update for all platforms but UWP. Looking at the .meta files, it appears none of them have UWP set. Can you take a pass, please?

david-rhodes:
I think this is good to merge? @brnkhy want to give a test?

BergWerkGIS:
Android and iOS looking good.

Editor and Windows not.

Editor has `Any CPU` and `Any OS` for both (x86 and x64) checked.

Windows also has both checked for sqlite@x86 and sqlite@64 leading to errors:
* `Plugin sqlite3.dll is used from several locations`
* `Plugins colliding with each other`


`sqlite/x86/sqlite3.dll`:
image
image

`sqlite/x64/sqlite3.dll`:
image
image

----

> Looking at the .meta files, it appears none of them have UWP set.

This is correct as UWP needs a different set of native sqlite binaries.
I've updated the docs in this branch to inlcude information about how to get sqlite into a `Windows Store` solution.

BergWerkGIS:
Tested with

* Windows native x86 ✅ 
* Windows native x64 ✅ 
* Android ✅ 

Check marks for iOs also looking good, haven't tested though.

Both x86 and x64 DLLs still have `Any CPU` and `Any OS` checked for `Editor`:
image
 but it worked without problems so I think we are good here - going to merge.

PR #145, opened 2017-07-06
Parent tiles to root, even on device. 

david-rhodes:
Removing editor specific parenting and positioning logic. This was intended as a performance optimization, but it is confusing and leads to bugs.


brnkhy:
@david-rhodes looks good to me 👍

PR #146, opened 2017-07-07
Earcut algorithm and performance improvements

brnkhy:
biggest perf improvements are;
-removed data string from feature behaviour monobehaviour. (they can be enabled for debugging by devs)
-changed texture modifier not to use mono behaviour. now it just sets the material without going through monobehaviour process. old texture modifier can be found under new name TextureMonobehaviourModifier
-extending lists are huge perf hit. always set capacity at initialization if you know it.


brnkhy:
unity_2017-07-08_01-35-57
found an issue at sutro tower

brnkhy:
unity_2017-07-08_01-58-14
probably a height mod bug in london

brnkhy:
unity_2017-07-08_02-11-23
rome, no side walls?

brnkhy:
unity_2017-07-08_02-13-35
little glitch in the middle of collesium

brnkhy:
sutro tower and colosseum errors looks like some lower level thing.
that building without side walls is bad data, it has min_height = 9, height = 9 hence flying plane

brnkhy:
just checked that Big Ben error (floating stuff in London), it looks exactly same in webgl so it's probably a data issue

sdkproject/Assets/Mapbox/Unity/MeshGeneration/Modifiers/MeshModifiers/HeightModifier.cs
```diff
@@ -79,18 +79,19 @@ public override void Run(VectorFeatureUnity feature, MeshData md, UnityTile tile
...
+			var count = md.Vertices.Count;
+			md.Vertices.Capacity = count + md.Edges.Count * 2;
+			float d = 0f;
+			Vector3 v1;
+			Vector3 v2;
+			int ind = 0;
 
-            var wallTri = new List<int>();
-            var wallUv = new List<Vector2>();
+			var wallTri = new List<int>(md.Edges.Count * 3);
```
david-rhodes: If we know the size, we should probably just use arrays. In fact, one future optimization might be to pool arrays in some memory manager singleton (referenced as `Scratch Arrays` here: https://youtu.be/F6vtgGY-aLs?t=31m20s).

sdkproject/Assets/Mapbox/Unity/MeshGeneration/Modifiers/MeshModifiers/ChamferHeightModifier.cs
```diff
@@ -0,0 +1,391 @@
...
+	public class ChamferHeightModifier : MeshModifier
+	{
+		[SerializeField]
+		private bool _flatTops;
+		[SerializeField]
+		private float _height;
+		[SerializeField]
+		private bool _forceHeight;
+		[SerializeField]
+		private float _offset;
```
david-rhodes: Let's add a range to this to prevent broken values!

sdkproject/Assets/Mapbox/Unity/MeshGeneration/Modifiers/MeshModifiers/ChamferHeightModifier.cs
```diff
@@ -0,0 +1,391 @@
...
+{
+	using System.Collections.Generic;
+	using UnityEngine;
+	using Mapbox.Unity.MeshGeneration.Data;
+
+	[CreateAssetMenu(menuName = "Mapbox/Modifiers/Smooth Height for Buildings Modifier")]
+	public class ChamferHeightModifier : MeshModifier
+	{
+		[SerializeField]
+		private bool _flatTops;
```
david-rhodes: if you don't check flat tops, the roof is not positioned correctly, relative to the walls.

PR #149, opened 2017-07-11
use string.Equals for testing mesh type filters

stevenschobert:
Hey folks! This is a quick fix for filters on mesh generation!

Currently, when adding a `TypeFilter` to a `VectorLayerVisualizer`, that filter will partially match types because the filter is using `string.Contains` for a check.

So if you add a filter to exclude `building:part`, this will end up excluding all `building` types in addition to `building:part`, because of the partially matching strings.

I'm not sure if that was the intended behavior or not, but since a `VectorLayerVisualizer` can have multiple filters, I don't think there is a strong reason to keep the "partial matching".

Let me know if I'm off-base, or can adjust anything!

---

By the way, thanks a ton for all the work going into this SDK. I've been working on a game with a few folks, and we're using it heavily! 🎉

### DIFF
--- a/sdkproject/Assets/Mapbox/Unity/MeshGeneration/Filters/TypeFilter.cs
+++ b/sdkproject/Assets/Mapbox/Unity/MeshGeneration/Filters/TypeFilter.cs
@@ -14,7 +14,7 @@ namespace Mapbox.Unity.MeshGeneration.Filters
 
         public override bool Try(VectorFeatureUnity feature)
         {
-            var check = _type.ToLowerInvariant().Contains(feature.Properties["type"].ToString().ToLowerInvariant());
+            var check = _type.ToLowerInvariant().Equals(feature.Properties["type"].ToString().ToLowerInvariant());
             return _behaviour == TypeFilterType.Include ? check : !check;
         }
 


brnkhy:
@stevenschobert ah that's an interesting case. The idea with that was to enable comma separated filter keys so that you can filter in/out stuff like "school,residential,commercial". 
It fails in your case as you said and I imagine there would be a few more interesting matches like that as well.
Maybe we can think of a better way to fix this, like instead of contains, filters may split by "," on enable and then check if that type is in the list.
also you got me curious, why do you need to filter out building parts? I'm asking because you'll lose lots of details/buildings in places like manhattan with that.

stevenschobert:
@brnkhy ah okay! Yeah I was wondering if that was the original goal, but since you can just supply an array of filters, I wasn't sure.

I'm happy to rework this PR into a better fix, if you want to go the string splitting approach?

I also was wondering if the `VectorLayerVisualizer`'s "stacks" property is using `Contains` under the hood too, but hadn't looked into it yet.

----

> also you got me curious, why do you need to filter out building parts? I'm asking because you'll lose lots of details/buildings in places like manhattan with that.

To be honest, I'm not 100% sure I do need to filter them (or what they're representing)!

I was running into a issue where the Unity Editor would hang, when trying to generate the buildings around DFW airport in Dallas. So I compared my setup against some default maps in Mapbox Studio, and saw they were filtering out building parts too. After I put in that filter in, it fixed the hanging issue.

I am forcing the height of all buildings to 0, could that cause issues with building parts?

brnkhy:
@stevenschobert I'm not sure how to change it to be honest, splitting comma separate values into an array is pretty much same solution and shouldn't be much slower either I guess.
Stack keys works same as you noted, we're thinking about replacing stack keys with filters though  (instead of string) so you'll have an option to put logic into that stack decision part. Unfortunately that requires lots of inspector work as well and not high priority at this time.

As far as I know, buıilding:part is used for subsections of big/detailed buildings but I'm not sure myself either. Can you give more details on which building is that, I can try to check. 
And instead of forcing height=0, you can simply remove height modifier from building stack.

stevenschobert:
Gotcha, I've removed the height modifier, thanks for the tip! 👍 

So in the editor I've got `32.896944444444, -97.038055555556` for the location, and the tiles immediately north are where it hangs (somewhere around terminal A)

screen shot 2017-07-11 at 2 45 15 pm

brnkhy:
@stevenschobert I cheated a little and tested with the latest stuff (that'll probably get merged in a day);
unity_2017-07-11_23-00-21
this "latest" version have a new algorithm for triangulation and bunch of bug fixes so it looks good on this. but please open tickets for any issues like this so we can test&fix them ;)

stevenschobert:
@brnkhy oh awesome! Can't wait to wire in the latest updates then! 👏 

Do you want me to close this PR for now? Or do you want to leave it open for discussion around handling the filters?

brnkhy:
@stevenschobert would you mind closing this and opening a ticket to talk about that key/contains issue? I think that would be better for tracking & discussion.

stevenschobert:
@brnkhy You bet! Opened up #150.

brnkhy:
@stevenschobert thanks!

PR #155, opened 2017-07-12
debugging tools and height mod wall option

brnkhy:


sdkproject/Assets/Mapbox/Unity/Editor/FeatureBehaviourEditor.cs
```diff
@@ -0,0 +1,25 @@
+﻿using UnityEngine;
+using System.Collections;
+using UnityEditor;
+using Mapbox.Unity.MeshGeneration.Components;
+
+[CustomEditor(typeof(FeatureBehaviour))]
```
david-rhodes: For something as simple as this, I recommend using [`ContextMenu` attribute](https://docs.unity3d.com/ScriptReference/ContextMenu.html). It's faster to write, leads to one less class, and keeps the inspector a bit cleaner.

brnkhy: But isn't that also harder to use compared to a huge button in the inspector? we'll eventually add more in that inspector anyway, that's the script keeping feature data on game objects so pretty much a core element.

PR #161, opened 2017-07-14
snap terrain branch

brnkhy:


sdkproject/Assets/Mapbox/Unity/MeshGeneration/Data/UnityTile.cs
```diff
@@ -121,6 +121,11 @@ internal void Recycle()
 			HeightDataState = TilePropertyState.None;
 			VectorDataState = TilePropertyState.None;
 
+			OnHeightDataChanged = null;
+			OnRasterDataChanged = null;
+			OnVectorDataChanged = null;
+			OnRecycled = null;
```
david-rhodes: This is not used. Do you think we need it?

sdkproject/Assets/Mapbox/Unity/MeshGeneration/Data/UnityTile.cs
```diff
@@ -121,6 +121,11 @@ internal void Recycle()
 			HeightDataState = TilePropertyState.None;
 			VectorDataState = TilePropertyState.None;
 
+			OnHeightDataChanged = null;
```
david-rhodes: We can't null these out, right? The empty delegate prevents need for null check on registration.
Also, what's the reason for this? We reset state, already.

brnkhy: it was supposed to be `delegate { };` , not null to clear events when the tile is recycled, to prevent events from previous tile of course.

PR #169, opened 2017-07-19
Tile shaving

david-rhodes:
Added new StyleOptimizedVectorTileFactory
MeshFactory is now VectorTileFactory (kept MeshFactory for backwards compatibility)
Style picker editor and search (hitting our API)--this is used for MapImageFactory custom style and new VectorTileFactory.
Create a style and add a single layer to test (building?).

For anyone interested, this is `Style-optimized vector tiles` here: https://www.mapbox.com/api-documentation/#retrieve-tiles (we filter out data you don't need for smaller file sizes).


brnkhy:
I couldn't test it properly (yet) but inspector ui looks quite confusing to me, we should work on that. 
First of all; Id, Name, Modified fields looks confusing. They are probably loaded from style so we can make them readonly. Also using an indentation and empty space before/after would help. Maybe with a smaller Clear button. I kinda like that R button for style reset for example.
Second, when I click Search and put username, it actually pops a nice error but it took me a little to understand it and explore api key page to see what it actually means. It would be great if we can put a big dialog there explaining it a little better with a link to a proper web page explaining what those scopes are. At least we should clearly point out; "you have an api key but this api key is create without this option and you need to create a new api key and check this option in creation window. then we'll be able to fetch our styles and list them here" etc.

david-rhodes:
@brnkhy Great feedback--thank you! I agree that the UI could use some polish. I did not mark these fields as `readonly` because I wanted developers to be able to use their current API keys (without creating a secret token just for `styles:list` support). In this way, you can continue to copy/paste the MapId just as we did previously (without ever using the search button). 

We should definitely explain the process better in the search window.

david-rhodes:
@brnkhy I think this is ready for review again :smirk:

brnkhy:
@david-rhodes I did some small changes, hopefully without breaking anything, to make it possible to save empty key. You can revert it if you don't like it or anything missing there but I like being able to clear the key, especially when you think about cases where we're doing spcial demos to someone and keep certain caching related values and not share our own keys etc.

PR #172, opened 2017-07-20
Fix for objectinspector modifier relying on example component

david-rhodes:
Moved objectinspector to examples folder, since it relies on other examples.
Fixed namespaces and references.

When removing `Examples` from the SDK, we got compile errors because of a missing reference. This should fix it.


brnkhy:
just changed font to arial as open sans didn't work for some reason (even though I set it initially). might be a OS (mac/windows) issue maybe? looks good to merge otherwise

Issue #162, opened 2017-07-14
Ensure tile IDs wrap at world boundaries correctly


david-rhodes:
`CameraBoundsTileProvider` and `CanonicalTileId` both suffer from an issue where the tiles do not correctly wrap at world boundaries (tile count max).

PR #175, opened 2017-07-21
Tiles should now wrap properly (and be placed accordingly).

david-rhodes:
TerrainFactories had improper direction based on Mapbox Tile Coordinates, so I fixed it.

As you approach the boundary of the world, tile IDs will properly wrap. Fixes: https://github.com/mapbox/mapbox-unity-sdk/issues/162.


brnkhy:
looks good to me. my only concern is that z axis (north/south) wrapping thing might be confusing and people would think it's broken. Not loading those tiles at all has more use cases I believe.

david-rhodes:
@brnkhy I agree, we could limit tiles for the Y-axis in the TileProvider, but then it might be confusing why the camera (by default) is allowed to roam outside the map bounds. Let's leave for now and revisit as needed.

Issue #101, opened 2017-05-30
Native plugin errors while building on Unity Cloud Build.


mihakrajnc:
I'm using UCB as my CI and both my Android and iOS builds are failing.

### iOS build is failing with the error:
```
[xcode] /BUILD_PATH/***********/temp.******/Libraries/Addons/Mapbox/Core/Plugins/iOS/MapboxMetricsClient.m:1:9: fatal error: 'MapboxMobileEvents/MapboxMobileEvents.h' file not found
6615:        [xcode] #import <MapboxMobileEvents/MapboxMobileEvents.h>
6616:        [xcode] 1 error generated.
```

I'm not an iOS developer, but looking at the `MapboxMetricsClient.m` file, i assume the import should be 
`#import <MapboxMobileEvents/include/MapboxMobileEvents/MapboxMobileEvents.h>`
and not
`#import <MapboxMobileEvents/MapboxMobileEvents.h>` ?

### Android build is failing with this error:
```
Exception in thread "main" java.lang.UnsupportedClassVersionError: com/android/apksigner/ApkSignerTool : Unsupported major.minor version 52.0
```

I didn't analyse the native `.aar` libraries that are used by Mapbox, but I'd say one of them is using Java 8 which UCB does not support.
Not sure if this is a problem that UCB should fix, or Mapbox.

isiyu:
thanks for flagging this @mihakrajnc - we will take a look on our end first. Any clues from your logs which `.aar` it's coming from? 

@boundsj @zugaldia - do you have ideas from the iOS / android errors?

@david-rhodes - can you recreate these errors by setting up cloud build for the slippy demo?

zugaldia:
> I didn't analyse the native .aar libraries that are used by Mapbox, but I'd say one of them is using Java 8 which UCB does not support.

The Android library we're using [explicitly sets the Java target to version 7](https://github.com/mapbox/mapbox-java/blob/master/mapbox/libandroid-telemetry/build.gradle#L15-L18). Could you see if adding the same lines to the Android app project helps with the issue?

```
compileOptions {
        sourceCompatibility JavaVersion.VERSION_1_7
        targetCompatibility JavaVersion.VERSION_1_7
    }
```

mihakrajnc:
@isiyu None, sorry, I checked the logs and all relevant information I get is this:

```
[...]

PR #179, opened 2017-07-24
Upgrade telem libraries

david-rhodes:
Fixes iOS-related issues in: https://github.com/mapbox/mapbox-unity-sdk/issues/101

Also adds method to enable/disable location collection, and hooked up to attribution buttons.


isiyu:
@david-rhodes  Seeing this on Android builds, loaded okay initially; but closed the app and  crashed on relaunch. 
hardware is LG G3; Android 5.0.1

screen shot 2017-07-25 at 4 17 37 pm

david-rhodes:
@isiyu before you get this exception, do you get any logs from Unity that look something like those that come from this method: https://github.com/mapbox/mapbox-unity-sdk/blob/f6aaef41844885b39bdafc13a28e05848f630731/sdkproject/Assets/Mapbox/Unity/Telemetry/TelemtryAndroid.cs#L20?

The error seems to suggest that `telemInstance` is null, so I would expect previous errors.

FWIW, I cannot reproduce on my Pixel.

/cc @zugaldia

BergWerkGIS:
Not able to repdocue on either of my devices: Android 7.0 and 7.1.1.

zugaldia:
The error seems to come from [this line](https://github.com/mapbox/mapbox-java/blob/master/mapbox/libandroid-telemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java#L601) when the `locationEngine` is `null`, which should never be the case as long as the module has been initialized with `initialize`. This could indicate an issue with the current set up in Unity, meanwhile I've ticketed https://github.com/mapbox/mapbox-java/issues/523 to avoid the exception and log an error message that could help us troubleshoot the problem.

isiyu:
@david-rhodes Looks like this may have been a bad build on my device. I deleted the app on the phone and did a fresh build and that seemed to have resolved it. (Built over the app with a few other demos using the same Package Name and all worked consistently.

PR #184, opened 2017-07-26
Example camera controller improvements

david-rhodes:
Added transform forward relative mouse zoom.
Handle case where camera is initially rotated (in Y) but panning works as expected.
Fixed issue where panning was dependent on aspect ratio (different horizontal/vertical speed).
Updated example scenes.


david-rhodes:
Not ready.

PR #185, opened 2017-07-27
preInit, Init and scaling

brnkhy:
@david-rhodes can you have a look at this, especially the preinit&init thing. No need to rush, let's not include this to this release. I just wanted to have your opinion

add PreInitialize and Initialize to mesh gen structure
add world properties class for shared data
add alternative world scaling based on scaling vertices rather than root game object


brnkhy:
@david-rhodes I'll push some fixes today, around/after scrum so you might want to skip this until then. I'll ping you again when it's ready to go.

brnkhy:
merged the dev into it as well, this should be ready for review but it won't be in the next release anyway so need to hurry for it.
@david-rhodes

Issue #136, opened 2017-06-23
Use unity-jar-resolver from google


MorganMoon:
It might be a good idea to use this for android:
https://github.com/googlesamples/unity-jar-resolver

It will minimize the risk of having multiple or conflicting versions of client libraries included in the Unity project.

Currently this sdk can have conflict issues for android

BergWerkGIS:
@MorganMoon are you experienced with `unity-jar-resolver`?

Looked into it, but it seems to me the only thing it does is collecting the libraries defined in c# code and put them into `Assets/PlugIns/Android`.

I still had to delete duplicate instances of libraries manually.
Or am I missing something?

BergWerkGIS:
@MorganMoon still fiddling with `unity-jar-resolver` but on the way I created a simple script that checks for duplicate libraries.

From a SDK user's perspective, what do you think, would that be enough to easily figure out why a build is failing?

![image](https://user-images.githubusercontent.com/4549888/27955951-9fac097e-6317-11e7-8c52-6c15f5cd2b47.png)

MorganMoon:
Hey, I am not very experienced to with it, I've just found it useful. The duplicate android plugins script looks like a nice addition as well :) If I remember right, I had removed most .aar plugins and let the unity-jar-resolver do it's thing, and everything seemed to work fine. However, I remember mapbox-android-telemetry-2-1-0.aar having some exceptions showing up in adb logcat. I am not currently using the most up do date mapbox sdk due to issue #101. I just found the unity-jar-resolver to pretty much allow mapbox android stuff to work with some other android stuff (facebook sdk) I am using. Thanks for looking into all of this!

BergWerkGIS:
@MorganMoon thanks for getting back.
I'll keep digging. 😄

BergWerkGIS:
Was not able to get `unity-jar-resolver` working reliably and consistently.
Tried several unitypackages, sources and the DLLs from the repo, but either libraries where not moved or not deleted or not picked up after a new asset import.
Always had to manually label them with `Gpsr`.

Ditching jar resolver for now and going to use the simple prebuild script described [above](https://github.com/mapbox/mapbox-unity-sdk/issues/136#issuecomment-313658150).

PR #186, opened 2017-07-28
check for duplicate Android libraries on prebuild step

BergWerkGIS:
Simple prebuild script that scans for duplicate Android libraries:

image

Maybe we revisit `Google jar-resolver` again later, refs #136


BergWerkGIS:
superseded by #191

PR #189, opened 2017-07-31
Building position

brnkhy:
@david-rhodes there's an option in the modifier stack (non-merged) to change center position. First option keeps it in tile center, second one uses first vertex (faster than average), third takes the average of points and uses that.


david-rhodes:
@brnkhy This breaks positioning for POIs (using `PrefabModifier`). Any way around this?

brnkhy:
@david-rhodes I pushed a one line to reset position, I'm not sure but I think it works. Was able to create broken scenes before it but not anymore. can you give it another shot?

PR #190, opened 2017-07-31
Camera control fixes

david-rhodes:
Improving our example camera movement script. This allows for a bit better panning on device.
Added mouse wheel zoom.


brnkhy:
loaded slippy, moved around zoomed in and out etc. there are bunch of ios related stuff in the same commit so it looks a little confusing but looks good to me 👍

PR #191, opened 2017-07-31
Check for Android library duplicates

BergWerkGIS:
This supersedes PR #186 and contains the pre-build check and an unit test for duplicate Android libraries.

image


david-rhodes:
@BergWerkGIS This is awesome--thanks! Unfortunately, `UnityEditor.Build` was only added in Unity 5.6. I've added an `#if` check, but I wonder if there's a more elegant way of handling this?

BergWerkGIS:
> Unfortunately, `UnityEditor.Build` was only added in Unity 5.6.

Thanks, completely missed that.

> I've added an `#if` check, but I wonder if there's a more elegant way of handling this? 

Hm, that disables the check for <5.6 completely.

Maybe add a custom build menu for `Unity<5.6`?
https://docs.unity3d.com/ScriptReference/BuildPipeline.BuildPlayer.html

PR #206, opened 2017-08-11
Node editor

brnkhy:
would love to take your feedback, even if we don't merge it right away

### DIFF
--- a/sdkproject/Assets/Mapbox/Unity/Map/AbstractMap.cs
+++ b/sdkproject/Assets/Mapbox/Unity/Map/AbstractMap.cs
@@ -44,6 +44,7 @@
 
 		[SerializeField]
 		MapVisualizer _mapVisualizer;
+		public MapVisualizer MapVisualizer { get { return _mapVisualizer; } }
 
 		[SerializeField]
 		float _unityTileSize = 100;
--- a/sdkproject/Assets/Mapbox/Unity/Map/MapVisualizer.cs
+++ b/sdkproject/Assets/Mapbox/Unity/Map/MapVisualizer.cs
@@ -20,6 +20,7 @@ namespace Mapbox.Unity.MeshGeneration
 	public class MapVisualizer : ScriptableObject
 	{
 		[SerializeField]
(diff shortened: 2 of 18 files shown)


brnkhy:
@david-rhodes ah alright totally missed that. any feedback on editor itself so far?

brnkhy:
as a note, I run into a strange behaviour where, if I'm not wrong, some scriptable object properties (like name) occsionally bugs out. I had this after moving the file but not running the project. I'm not sure if it's that yet, but everything works properly once you run the project.

david-rhodes:
@brnkhy some notes:
- Moving the window around causes nodes to disappear or shift. It appears to be some sort of double transform. The `reset zoom` does not appear to work.
- The graph has some clipping issues where the node rendering suddenly cuts off even though the nodes are well inside the graph window. This appears to be related to zoom.
- Should `Reset zoom` attempt to fit all nodes  into the current window size? If so, this does not work, currently. Additionally, I feel that reset zoom should also reset the "pan." Consider this a "frame all" button?
- The menu item to open this editor should be nested under `Mapbox` menu
- Instead of selecting which `MapVisualizer` you want to render the graph for, why not update the graph when a new `MapVisualizer` scriptable object is selected? Always render the most recent selected asset. As an option, the current dropdown you have could show all `MapVisualizers` in the project (versus the current scene).

brnkhy:
@david-rhodes 
Fixed last two points (and folders&namespaces) already. 2nd and 3rd is all about zoom thing, draw rect bugs out when zoom level != 1. I like the idea to reset pan with zoom, will add that right away.
Zoom isn't supposed to fit graph, just set zoom level to 1. Fitting thing is a good idea actually and I'm pretty sure I can do that, but I'll have to fix that zoom bug first I guess.

I don't get the first one though, you mean moving the node editor window? Tried a few different things but couldn't get anything like that, can you elaborate?

brnkhy:
I named it "Map Editor" by the way, let me know if you have better ideas.

brnkhy:
another note; missing objects breaks the editor, should handle that properly.

PR #215, opened 2017-08-22
Node editor folder restructure

brnkhy:


brnkhy:
`- in the inspector for a modifier should also delete the entry in the list, right?`
not sure what this means but I noticed "-" thing clears the field (default behaviour when you remove something from list) for some scripts and removes it totally for others (running removal code twice, strance).  First one has side effects though, I'll try to do it properly but clearing field at first click is ok as long as we have null checks there I guess (we don't, yet)

`required references ("missing")`
probably messed up/forgot some meta file during commit.

`Can we clarify that this represents some sort of asset picker?`
I'll try to change this with a magnifier icon I guess.

`Optimized vector tile factory is missing custom editor`
totally missed that, will check it.

Also found a bug with "add new button" and vector layer visualizer. checking them at the moment.

david-rhodes:
@brnkhy Did some additional cleanup and example scene repair. Only thing left I noticed is that non-default stacks do not render modifiers. Is this by design?

Anyway, please make sure I didn't break anything by moving scripts and changing namespace names.

screen shot 2017-08-30 at 2 23 45 pm

brnkhy:
@david-rhodes ` Is this by design?` definitely not, not sure what's wrong there but checking at the moment.

brnkhy:
@david-rhodes I fixed that problem. Turns out it's introduced after that custom stack keys addition.
unity_2017-08-31_18-07-50

PR #230, opened 2017-09-05
Mesh.GetVertices and Mesh.GetNormals do not exist in Unity 5.5.0 and 5.5.1

mikelortega:
Mesh.GetVertices and Mesh.GetNormals were [introduced in Unity 5.5.2](https://unity3d.com/es/unity/whats-new/unity-5.5.2) so Versions 5.5.0 and 5.5.1 should be excluded.


david-rhodes:
Thanks @mikelortega!

PR #240, opened 2017-09-14
Slippy map with dynamic zoom

BergWerkGIS:
1st iteration of a zoomable slippy map implementation.

PR to discuss general design and concrete implementation:

* effects on other examples
* effects on SDK

known issues:
* node editor is broken


david-rhodes:
@BergWerkGIS back to you for counter-review. Please make sure I didn't break anything and that you are OK with the changes and the design makes sense.

A few things:
- https://github.com/mapbox/mapbox-unity-sdk/pull/240/commits/6e3a893108209fdcd3047294dab6b316b9efc7c0 this `OriginShift` value appears to be wrong. I first noticed with the `DirectionsFactory` (`Drive` example directions were wrong)
- There is still some work left, but this is the bulk of it
- Still need to validate node editor and such

BergWerkGIS:
thanks @david-rhodes, looks good, nice loading indicator 👍 

----

> this OriginShift value appears to be wrong. I first noticed with the DirectionsFactory (Drive example directions were wrong)

Then we have to revisit the `Drive` example.

Using `48.2072,16.421` which is in the middle of the `Ernst Happel Stadion`
https://www.openstreetmap.org/?mlat=48.20718&mlon=16.42096#map=18/48.20718/16.42096
with `OriginShift = 20037508.34` DynamicZoom example starts at the right location:

image

whereas using the formula puts the map quite a few kms too far south and a bit to the west.

Can be verified via the debug messages and an online converter, eg http://tool-online.com/en/coordinate-converter.php

image

with `20037508.34`: ✅ 
image

with formula: :x:
image

--- 

Other TODOs 

* panning doesn't seem to work as expected anymore

david-rhodes:
@BergWerkGIS Once you're done investigating `Conversions`, can you please remove dead comments and logs (or wrap them in preprocessor directives)?

@brnkhy Can you take a look at this branch and help us fix the node editor, please?

brnkhy:
@david-rhodes @BergWerkGIS 
checked the node editor thing, I think only problem is that the factories inside AbstractMapVisualizer is private and node editor can only fetch public fields or properties at the moment. Changing that (AbstractMapVisualizer line 16 to public) fixes the issue. 
We can do that (I'm not pushing that yet) or we can try to change reflection stuff in node editor to fetch private fields I guess. I think factories of a map visualizer should be public though (same with active tiles property but that's a whole different topic)

Issue #194, opened 2017-08-02
Factory/MapVisualizer events are not working as intended with caching


brnkhy:
This is a problem with Factory and MapVisualizer state change events and how they behave with cached data.

In our current structure MapVisualizer registers tiles to factories one by one. Then factories make a web call for the relevant tile data and process the tile afterward.
We did a basic progress tracking at this point by; 
- increment `progress` on tile registration
- fetch tile data and process it
- decreament `progress`

and on the get/set of the `progress` value, we fire state change event if progress goes from zero to 1 (loading) and 1 to 0 (finished).

Go in general it goes like this; (numbers are `progress` value)
`1(working event) 2 3 4 3 2 1 0 (finished event)`

Important thing here is that, web calls are done in a coroutine so system actually moves to next tile while waiting for the response. So all those increments and decrements are in different frames.

But with caching, there's no web call delays and no coroutines anymore so that sequence turns into;
`1(working event) 0(finished event) 1(working event) 0(finished event) 1(working event)... `

I realized this working on an urgent demo so my quick fix was adding a coroutine after `progress` increment and push everything afterwards one frame which gave time to register all tiles before start processing them. This might sound like a hacky solution but technically it's saying; "let's gather all requests this frame, know our workload and then process them next frame", which makes sense to me. I'll post more on this as a response below.

Another solution might be, taking requests as list instead of one by one. That's not bad either considering that even one request can be passed as a list of one (wasteful?).

Obviously fire those events and doing it in getter/setter isn't really a good idea but I don't think changing that alone would fix this issue as the main cause, factory not knowing its workload, will still be an issue.

@david-rhodes @isiyu

brnkhy:
About that coroutine solution, there's one obvious flaw with that; what happens there's another request in that second frame. 
I haven't tried this so I'm just speculating but I believe that'll also break it and it'll fire finished event twice again. Using a flag to detect if there are any new tile requests in that frame would help. Then coroutine can check that and wait yet another frame to process. Or we can have a queue or something and process it on empty frames etc.
That feels theoretically wrong though, as it might mean waiting forever as well (not that it'll ever happen practically).
We can delay the event one frame. That would solve this all (with checking if progress is still 0 next frame of course) but then we'll have issues where devs expecting the event in that frame (as it really finished) and not getting it etc.
We can fix this with bunch of flags and checks really but I don't want to go into that.

NinjaDanz3r:
[...]

PR #259, opened 2017-09-26
Delayed cache response

david-rhodes:
Delay any caches response by one frame to allow other systems to initialize before responses are available.

Fixes: https://github.com/mapbox/mapbox-unity-sdk/issues/194


david-rhodes:
Pre-approved by @brnkhy and @BergWerkGIS

PR #292, opened 2017-10-16
Performance Improvements

brnkhy:


sdkproject/Assets/Mapbox/Unity/MeshGeneration/Modifiers/GameObjectModifiers/PrefabModifier.cs
```diff
@@ -14,22 +14,22 @@ public class PrefabModifier : GameObjectModifier
         [SerializeField]
         private bool _scaleDownWithWorld = false;
 
-        public override void Run(FeatureBehaviour fb, UnityTile tile)
+        public override void Run(VectorEntity ve, UnityTile tile)
         {
-            int selpos = fb.Data.Points[0].Count / 2;
-            var met = fb.Data.Points[0][selpos];
+            int selpos = ve.Feature.Points[0].Count / 2;
```
david-rhodes: Poi demo appears to be breaking here. . .

sdkproject/Assets/Mapbox/Unity/Utilities/VectorEntity.cs
```diff
@@ -0,0 +1,15 @@
+using Mapbox.Unity.MeshGeneration.Components;
```
david-rhodes: Now that we have vector entity, and it contains `Mesh`, can we streamline the interface for `Modifiers`? In other words, mesh and game object modifiers receive the same params?

sdkproject/Assets/Mapbox/Unity/Editor/VectorTileFactoryEditor.cs
```diff
@@ -9,12 +9,15 @@
 	[CustomEditor(typeof(VectorTileFactory))]
 	public class VectorTileFactoryEditor : UnityEditor.Editor
```
david-rhodes: Don't forget about `StyleOptimizedVectorTileFactoryEditor.`

sdkproject/Assets/Mapbox/Unity/MeshGeneration/Modifiers/ModifierStack.cs
```diff
@@ -21,147 +20,155 @@ public enum PositionTargetType
...
+			_tempVectorEntity = _pool.GetObject();
+			_tempVectorEntity.GameObject.SetActive(true);
+			_tempVectorEntity.Mesh.Clear();
 
-			ve.GameObject.name = type + " - " + feature.Data.Id;
-			ve.Mesh.subMeshCount = meshData.Triangles.Count;
-			ve.Mesh.SetVertices(meshData.Vertices);
-			ve.Mesh.SetNormals(meshData.Normals);
+			_tempVectorEntity.GameObject.name = type + " - " + feature.Data.Id;
+			_tempVectorEntity.Mesh.subMeshCount = meshData.Triangles.Count;
```
david-rhodes: Cache mesh and then set the property to avoid property access cost.

sdkproject/Assets/Mapbox/Unity/MeshGeneration/Modifiers/ModifierStack.cs
```diff
@@ -21,147 +20,155 @@ public enum PositionTargetType
...
+				_activeObjects.Add(tile, _listPool.GetObject());
+			}
+			_activeObjects[tile].Add(_tempVectorEntity);
 
 
-			ve.Transform.localPosition = _center;
+			_tempVectorEntity.Transform.localPosition = _center;
 
-			foreach (GameObjectModifier mod in GoModifiers.Where(x => x.Active))
+			for (int i = 0; i < GoModifiers.Count; i++)
```
david-rhodes: Cache `Count` in situations like this.

david-rhodes:
We should merge develop in soon.

brnkhy:
@david-rhodes 
`Panning to the ocean in MeshGeneration scene will show buildings in water.`
been looking into this for hours now. It only happens when you move camera crazy fast and start/destroy 4-5 tiles in a single frame. Couldn't fix it, I'm not even sure where the problem is. I think there's a problem with stopping coroutines, might even be a problem with Runnable class, I'm not sure.

`We should streamline modifier parameters, now (abstract class)?`
You mean mesh&go mods getting same parameters? I think so, I'll check this.

`With FeatureBehaviour gone, what is the recommended method of fetching feature data as a post map construction process?`
I think it's still adding that monoscript, just through another game object modifier.

sdkproject/Assets/Mapbox/Unity/MeshGeneration/LayerVisualizers/VectorLayerVisualizer.cs
```diff
@@ -50,12 +51,20 @@ public override string Key
...
 		[NodeEditorElementAttribute("Custom Stacks")]
 		public List<TypeVisualizerTuple> Stacks;
-		
-		private GameObject _container;
+
+		[NonSerialized]
+		private Dictionary<UnityTile, List<int>> _activeCoroutines;
+
+		[NonSerialized]
+		private int _entityPerCoroutine = 20;
```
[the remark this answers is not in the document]
brnkhy: I think I was changing that param back&forth while testing and I generally put that nonserialized thing to private field these days as I had some crazy bugs because of that.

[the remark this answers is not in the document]
brnkhy: we can remove&public that when we finalize this by the way

brnkhy:
merged latest on dev

david-rhodes:
For posterity: 
- cache count property (in both Mod stacks and anywhere else you see it in a for loop)
- expose # objects to process for coroutines
- streamline modifiers 
- map visualizer state is still broken (as a result of disk caching)
- does changing a scene during processing break everything?

PR #300, opened 2017-10-18
Added check to latitude/longitude coordinates

thedrummeraki:
I thought of adding something very simple that's been bugging me for a while. 

Since my laptop's locale is not English at home, I have to change `AbstractMap` when trying to generate a map every to adapt to my locale. So I added a check that basically that basically checks if the coordinates are in the format `0.000, 0.000`. 

This format makes sense for most coordinates (I tried `0,000, 0,000` and it looks very awkward).

### DIFF
--- a/sdkproject/Assets/Mapbox/Unity/Map/AbstractMap.cs
+++ b/sdkproject/Assets/Mapbox/Unity/Map/AbstractMap.cs
@@ -1,6 +1,7 @@
 ﻿namespace Mapbox.Unity.Map
 {
 	using System;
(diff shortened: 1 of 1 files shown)


david-rhodes:
@thedrummeraki Awesome--thank you! I can't modify your PR, so I recreated here with your code as a base: https://github.com/mapbox/mapbox-unity-sdk/pull/312.

I figured this would be more modular. That said, I don't have much experience with locale and all that. Can you verify that it works as expected, please?

PR #302, opened 2017-10-19
Mapbox Map Matching API

BergWerkGIS:
* Implements the Map Matching API https://www.mapbox.com/api-documentation/#map-matching
* Adds some missing properties to `Directions` classes
* Has tests :smile:

image

### DIFF
--- a/sdkproject/Assets/Mapbox/Core/mapbox-sdk-cs/MapMatching/MapMatcher.cs
+++ b/sdkproject/Assets/Mapbox/Core/mapbox-sdk-cs/MapMatching/MapMatcher.cs
@@ -57,11 +57,11 @@ namespace Mapbox.MapMatching
 
 
 		/// <summary>
-		/// Deserialize the geocode response string into a <see cref="GeocodeResponse"/>.
+		/// Deserialize the map match response string into a <see cref="MapMatchingResponse"/>.
 		/// </summary>
 		/// <param name="str">JSON String.</param>
-		/// <returns>A <see cref="GeocodeResponse"/>.</returns>
-		/// <typeparam name="T">Forward or reverse geocode. </typeparam>
+		/// <returns>A <see cref="MapMatchingResponse"/>.</returns>
+		/// <typeparam name="T">Map Matcher. </typeparam>
 		internal T Deserialize<T>(string str)
 		{
 			return JsonConvert.DeserializeObject<T>(str, JsonConverters.Converters);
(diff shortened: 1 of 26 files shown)


sdkproject/Assets/Mapbox/Examples/LocationProvider/Scripts/MapMatchingExample.cs
```diff
@@ -0,0 +1,79 @@
...
+				Debug.LogError("MapMatchingExample: " + response.MatchingError);
+				return;
+			}
+
+			_lineRenderer.positionCount = 0;
+			for (int i = 0, responseTracepointsLength = response.Tracepoints.Length; i < responseTracepointsLength; i++)
+			{
+				var point = response.Tracepoints[i];
+
+				// FIXME: why/how can a point be null?
```
david-rhodes: @BergWerkGIS Any idea why I have to check if these points are null? How is that possible?

BergWerkGIS: https://github.com/mapbox/mapbox-unity-sdk/blob/map-matching/sdkproject/Assets/Mapbox/Core/mapbox-sdk-cs/MapMatching/MapMatchingResource.cs#L127-L128

https://github.com/mapbox/mapbox-unity-sdk/blob/map-matching/sdkproject/Assets/Mapbox/Core/mapbox-sdk-cs/Tests/UnitTests/Editor/MapMatcherTest.cs#L467

sdkproject/Assets/Mapbox/Core/mapbox-sdk-cs/MapMatching/MapMatcher.cs
```diff
@@ -0,0 +1,70 @@
...
+					{
+						data.SetRequestExceptions(response.Exceptions);
+					}
+
+					callback(data);
+				});
+		}
+
+
+		/// <summary>
```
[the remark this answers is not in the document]
BergWerkGIS: good catch, thanks.

sdkproject/Assets/Mapbox/Examples/LocationProvider/Scripts/MapMatchingExample.cs
BergWerkGIS: Also:
> If a trace point is omitted by map matching because it is an outlier, the entry will be null.

https://www.mapbox.com/api-documentation/#match-response-object

david-rhodes: Thanks for the clarification. I read that, but assumed those entries would have simply been excluded from the server (and I wasn't using `Tidy`). Strange.

BergWerkGIS: To be able to evaluate results with regards to input data all arrays (coordinates, tracepoints, radiuses, timestamps, ...) of requests and responses have to be of same length always.

Issue #239, opened 2017-09-14
Error on Hololens (Screen position)


taszilvo:
Hi.

After i build and try to deploy on Hololens device the example scene (MeshGenerationBasic), before the Unity splash screen my app just stopped working. (Ofc i did the options for UWP and Hololens as in the description).
The error message is: "Screen position out of view frustum (screen pos 0.000000, 0.000000) (Camera rect 0 0 1268 720)"

I use: Unity 5.6.0.f3 and visual studio 2015.  (Mapbox is not working in 2017)

david-rhodes:
@taszilvo >Mapbox is not working in 2017
For Hololens, or in general? Can you elaborate, please?

@BergWerkGIS have you ever seen this? Any ideas?

taszilvo:
I tried only for Hololens (UWP project).
The generated project by Unity5.6.0f3 build successful in VS 2015, but in 2017 i cant build, get this error at building the UWP project:
[https://ibb.co/f7G3w5](url)

BergWerkGIS:
As for the error: I don't have any ideas.

As for VS2017: 
* generally speaking the SDK does work with VS2017 - I'm even on ` VS2017 15.4.0 Preview 2.0`
* I ran into a different problem before: I couldn't build HoloLens projects with VS2015 but only with VS2017.
In your case I think it boils down to Unity not being able to work with the W10/UWP SDK you have installed - the one coming with VS2017 probably being too new.

taszilvo:
But i think, you tried Mapbox on Hololens (or just emulator) successfully (because you made a tutorial to Hololens).  But how? :D    Unity version, VS version, etc..

BergWerkGIS:
@taszilvo I don't have a physical HoloLens to test.

This combination works for me with `Mixed Reality Portal`:

* Unity 2017.1.1f1 
* VS2017 15.4.0 Preview 2.0
* WinSDK 10.0.16278.0

Unfortunately there's an error with  `HoloLens Emulator 10.0.14393.0`:
`The program has exited with code -1073741819 (0xc0000005) 'Access violation'.`

Currently I don't have any idea where and why this is happening.
We'll have to do keep digging.

taszilvo:
Yes, i have HoloLens Emulator 10.0.14393.0
I tried with newer Unity and VS versions too, and i got this error too...
So that's the problem. :(

jordy-isaac:
If you're still having trouble with version 1.4 please re-open this ticket.

PR #306, opened 2017-10-20
"Mapbox.IO.Compression-unity" binaries

### LINKED ISSUE #284: Provide DLLs for Mapbox.IO.Compression-unity
Dependency [`Mapbox.IO.Compression-unity`](https://github.com/mapbox/Mapbox.IO.Compression-unity) does not change often.
Provide binaries to be included with the SDK to reduce number of included source files and Unity compilation times.

BergWerkGIS:
* closes #284 
* removes submodule `dependencies/Mapbox.IO.Compression-unity`
* adds assemblies (DLLs) for Net3.5 and UWP
* tested on
  * ✅  Windows native x64
  * ✅  Android
  * UWP: problems see below
------

When testing this branch make sure to run
* Windows: `update-mapbox-unity-sdk-core.bat`
* Linux/OSX: `./update-mapbox-unity-sdk-core.sh`

after checkout to get rid of the submodule and get the new DLLs.

**And also when checking out another branch afterwards!**

------

### Having troubles with the `meta` files and UWP builds.

Using these settings :
* Net35 DLL (`Any Platform` / `Exclude Platforms`):
  image
* UWP DLL (`Include Platforms`):
  image

and exporting to `Windows Universal Platform` I get error `'dll' is used from several locations`:

image

**Why?**

-------

When explicitly stating `Include Platforms` for both DLLs

image image

export works expected.

But that's not what I want as being on Windows I cannot set properties eg for `iOS` platform.


david-rhodes:
@BergWerkGIS For some reason, Unity only exposes the platforms that you currently have modules installed for. The best approach is likely to copy properties directly from the JSON plugin meta files, as those have already been sorted (manually).

BergWerkGIS:
@david-rhodes 

> For some reason, Unity only exposes the platforms that you currently have modules installed for.

Yes, but isn't that a bug in Unity?

Why does the **`Net35`** assembly get included (`'dll' is used from several locations`) when exporting to **`UWP`** using these settings on it:
*  `Any Platform` / `Exclude Platforms`
  image

As I understand it, it should be used for **all** platforms _**except**_ `UWP`, no?

BergWerkGIS:
Yes, indeed a Unity bug, which should be solved with `2017.3.0a7`:
https://issuetracker.unity3d.com/issues/dll-files-are-not-excluded-during-building-for-standalone-platform-even-though-they-are-selected-to-be-excluded

I was able to reproduce that issue being solved with `2017.3.0.b5`.

I suppose for the time being we have to stick with explicit `include` settings.
Will update using `meta`s from `Mapbox.Json` soon.

BergWerkGIS:
Ready for review.

Please test on as many platforms as possible: OSX, Windows, iOS, Andoird.

david-rhodes:
- [ ] iOS (6s+, iOS 11)
- [ ] Android (Pixel)
- [ ] WebGL 
- [ ] OSX standalone

david-rhodes:
@BergWerkGIS You tested UWP and all is good?

BergWerkGIS:
Exporting UWP project and building that works.
Deploying the built UWP app to the HoloLens Emulator still crashes on my machine, see #239

PR #317, opened 2017-10-24
Implemented tile error callbacks in AbstractTileFactory and AbstractMapVisualizer

apavani:
@david-rhodes 
Can you please check the error callback implementation. I'm just starting a pull request to start a discussion. I may have to do more changes on it based on your review.


BergWerkGIS:
I suggest that events pass more information than just the tile id.
Tile id might not enough to quickly figure out what went wrong: could be an error during the web request or during parsing.

I don't want to be counter productive, just asking:
Why do we need error callbacks at all?

Base class `Tile.cs` has [request error information attached](https://github.com/mapbox/mapbox-unity-sdk/blob/448ce2106ce0a296f0b4527c4a5164c37a6fdec0/sdkproject/Assets/Mapbox/Core/mapbox-sdk-cs/Map/Tile.cs#L55-L79).
Derived classes (eg `VectorTile.cs`) augments that [information with parsing errors](https://github.com/mapbox/mapbox-unity-sdk/blob/448ce2106ce0a296f0b4527c4a5164c37a6fdec0/sdkproject/Assets/Mapbox/Core/mapbox-sdk-cs/Map/VectorTile.cs#L160).

`UnityTile` already has property `VectorDataState`.

Additionally exposing `Tile.cs`'s properties `HasError`, `Exceptions` and `ExceptionsAsString` in `UnityTile` would allow all client code to check for errors:

eg. `AbstractMapVisualizer.cs` would just have to check those properties [here between `unityTile.Initialize()` and `PlaceTile()`](https://github.com/mapbox/mapbox-unity-sdk/blob/448ce2106ce0a296f0b4527c4a5164c37a6fdec0/sdkproject/Assets/Mapbox/Unity/Map/AbstractMapVisualizer.cs#L149-L150).

No need to complicate, or am I missing something?

BergWerkGIS:
When adding new functionality please also add tests:
* `mapbox-sdk-cs` related: [`sdkproject/Assets/Mapbox/Core/mapbox-sdk-cs/Tests/UnitTests/Editor`](https://github.com/mapbox/mapbox-unity-sdk/tree/develop/sdkproject/Assets/Mapbox/Core/mapbox-sdk-cs/Tests/UnitTests/Editor)
* Unity related: [`/sdkproject/Assets/Mapbox/Unity/Editor/Tests`](https://github.com/mapbox/mapbox-unity-sdk/tree/develop/sdkproject/Assets/Mapbox/Unity/Editor/Tests)

david-rhodes:
@BergWerkGIS Agreed--we definitely need more than the `TileId`. The initial idea was to have the tile factories propagate any errors as an event (containing TileId and original exception list). We cannot simply check properties in `AbstractMapVisualizer` because these requests are asynchronous and we won't yet have any errors logged.

That said, `Tile` needs no modification, as presented here. These errors are already available in various objects, as @BergWerkGIS pointed out. I like the idea of possibly capturing errors in `UnityTile` as well, but how would a client easily get access to a specific failed tile? The most likely use-case for this is requesting a tile that failed for some reason (time out, rate request limit exceeded, etc.). In this case, having an exception handler object listening to tile factories (or even the map visualizer) for errors (as events) seems the most useful.

So, in my mind, all we have to do is "centralize" a place where clients can listen for any tile error. AbstractTileFactory is the only place we currently have (a reference to) where we know the exact moment a tile request has returned (and whether it has an error or not). We can easily bubble that to `AbstractMapVisualizer` as well. Doesn't seem complicated to me, but if you have a simpler suggestion, I'm happy to hear.

sdkproject/Assets/Mapbox/Unity/Map/RangeTileProvider.cs
```diff
@@ -19,11 +19,14 @@ public override void OnInitialized()
 			var centerTile = TileCover.CoordinateToTileId(_map.CenterLatitudeLongitude, _map.Zoom);
 			AddTile(new UnwrappedTileId(_map.Zoom, centerTile.X, centerTile.Y));
 
-			for (int x = (int)(centerTile.X - _west); x <= (centerTile.X + _east); x++)
+			for (int i= 0; i < 10000; i++)
```
david-rhodes: Did you intend to leave this here? I assume this was to force errors?

sdkproject/Assets/Mapbox/Unity/MeshGeneration/Data/UnityTile.cs
```diff
@@ -230,23 +230,23 @@ public Texture2D GetRasterData()
...
 		}
 
         //OnTileError delegate handle method to bubble up the event all the way up to the chain to AbstractTileFactory and MapVisualizer
-        private void OnTileError(CanonicalTileId id){
-            if (OnTileErrorEvent != null)
+		private void OnTileErrorHandler(TileErrorEventArgs e){
+            if (OnTileError != null)
             {
-                OnTileErrorEvent(id);
+				OnTileError(new TileErrorEventArgs(e.TileId,this,e.Exceptions));
```
david-rhodes: Because event properties are private set, you cannot simply modify existing event args instance. This leads to additional garbage. I'm wondering if encapsulation for event args offers many benefits?

sdkproject/Assets/Mapbox/Examples/Scripts/TileErrorHandler.cs
```diff
@@ -0,0 +1,28 @@
...
+
+/// <summary>
+/// Monobehavior Script to handle TileErrors. 
+/// There's an OnTileError event on AbstractMapVisualizer, AbstractTileFactory and UnityTile classes that one can subscribe to to listen to tile errors
+/// </summary>
+public class TileErrorHandler : MonoBehaviour {
+
+
+	void OnEnable () {
+		GetComponent<AbstractMap>().MapVisualizer.OnTileError += _OnTileErrorHandler;
```
david-rhodes: Prefer `[SerializeField`] for additional developer flexibility--only get component if assigned reference is null.

sdkproject/Assets/Mapbox/Examples/Scripts/TileErrorHandler.cs
```diff
@@ -0,0 +1,28 @@
...
+	}
+
+	private void _OnTileErrorHandler(Mapbox.Map.TileErrorEventArgs e)
+	{
+		foreach (var exception in e.Exceptions)
+		{
+			Debug.Log(exception);
```
david-rhodes: Let's also include tile id here?

sdkproject/Assets/Mapbox/Core/mapbox-sdk-cs/Map/TileErrorEventArgs.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace Mapbox.Map
+{
+	using System;
+	using Mapbox.Unity.MeshGeneration.Data;
+	using System.Collections.Generic;
+
+	public class TileErrorEventArgs {
```
david-rhodes: Does it make sense to include type type, as well? If a tile request fails, how would we issue a specific request again?

sdkproject/Assets/Mapbox/Examples/Scripts/TileErrorHandler.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System.Collections;
```
david-rhodes: let's put this in the examples namespace and rename the class to make it obvious this is an example of how to register for these events.

sdkproject/Assets/Mapbox/Core/mapbox-sdk-cs/Map/TileErrorEventArgs.cs
apavani: Done!

sdkproject/Assets/Mapbox/Examples/Scripts/TileErrorHandler.cs
apavani: How does HandleTileErrorsExample sound as a class name?

david-rhodes: Sounds great!

sdkproject/Assets/Mapbox/Examples/Scripts/TileErrorHandler.cs
apavani: Done!

sdkproject/Assets/Mapbox/Examples/Scripts/TileErrorHandler.cs
apavani: Done!

sdkproject/Assets/Mapbox/Unity/Map/RangeTileProvider.cs
apavani: Done!

apavani: Good catch. Forgot to take that line out

sdkproject/Assets/Mapbox/Unity/MeshGeneration/Data/UnityTile.cs
apavani: While it doesn't offer much benefits beyond just preventing classes from randomly garbling the data on an existing TileErrorEventArgs instance.
But, I guess that is what we need here anyways. I'll make them public get; set;

apavani: Another question is, do we need a unitytile instance on the errorevent?

david-rhodes: Hmm, may be needed for some types of exception handling? Perhaps injecting data into the UnityTile instance after making a new request outside factory framework?

sdkproject/Assets/Mapbox/Unity/Utilities/HTTPRequest.cs
```diff
@@ -57,26 +55,13 @@ public void Cancel()
...
-			{
-				yield return null;
-				if (DateTime.Now > timeout)
-				{
-					_request.Abort();
-					didTimeout = true;
-					break;
-				}
-			}
+			yield return _request.SendWebRequest();
```
david-rhodes: I think this is only 2017.2+? Getting missing assembly reference in 2017.1.

sdkproject/Assets/Mapbox/Unity/Utilities/HTTPRequest.cs
```diff
@@ -57,26 +55,13 @@ public void Cancel()
...
 
 			Response response;
-			if (didTimeout)
+			if (_request.isHttpError || _request.isNetworkError)
```
david-rhodes: Do we need an if check anymore?

BergWerkGIS: Yes, I think so. Or refactor the following lines a bit to account for cancelled requests.

sdkproject/Assets/Mapbox/Unity/Utilities/HTTPRequest.cs
BergWerkGIS: Sorry, my bad. Thought it's 2017.1 already.
Just checked Unity docs, it isn't.

sdkproject/Assets/Mapbox/Unity/Utilities/HTTPRequest.cs
david-rhodes: `response = Response.FromWebResponse(this, _request, _wasCancelled ? new Exception("Request Cancelled") : null);` 

This covers all cases, right?

david-rhodes:
@apavani we need to roll back changes to `HttpRequest` (at least the API breaking changes) and update inline comments to reflect the new event args, please.

Thanks!

apavani:
Just reverted back @BergWerkGIS 's commit to `HTTPRequest` and also added / modified code comments. Ready for another review @david-rhodes

david-rhodes:
@apavani Interesting issue I just found: if you cancel a tile request, you will not receive any exceptions (which would normally include an `aborted` exception. This is because `UnityTile` unregisters from the tile exception event when cancel is called. 

Should we instead unregister in `Recycle`?

apavani:
@david-rhodes I see that Recycle() calls Cancel(). So, isn't unregistering in Recycle mean the same thing as unregistering in Cancel(). Would it make more sense to Unregister in OnDestroy()?
Also, just for my knowledge how do I cancel a tile request, just by calling Cancel()?

david-rhodes:
As per voice, let's note this "aborted" exception in [known issues](https://www.mapbox.com/mapbox-unity-sdk/docs/02-known-issues.html). But, since a cancel is issued in client code, this is not really an exception case (no errors will result from this).

david-rhodes:
@BergWerkGIS Want to do a final review before we merge?

BergWerkGIS:
I know this comes a bit late but does it make sense to explicitly have to subscribe to errors?
I was pleasantly surprised to not get any errors at all with `Globe@z6` and it looked great at first glance.
Only to find out it looked liked this when turning:
image

Stupid me, but I forgot to attach `HandleTileErrorExample`.

From my point of view, the error handler should be attached to the map object per default.

Trying to put myself into the shoes of a dev new to the SDK:
Tiles don't display and I don't know why? There are no errors.
On the other hand, at least I know what's wrong and where to start looking.
And if I choose to unsubscribe I am able to do so and just ignore errors.

----

Please derive `TileErrorEventArgs` from `EventArgs`.
It's good practice and it is necessary for the next steps.

-----

Not sure if doing this is absolutely necessary with our current implementation as it's mainly needed when multithreading is involved but nevertheless I think it's good practice to always work on a reference of the event handler like this:

Taken from example at [`Microsoft Docs, How to: Raise and Consume Events`](https://docs.microsoft.com/en-us/dotnet/standard/events/how-to-raise-and-consume-events):
Instead of
```c#
if (ThresholdReached != null)
{
  ThresholdReached(this, e);
}
```
do
```c#
EventHandler handler = ThresholdReached;
if (handler != null)
{
  handler(this, e);
}
```

After deriving from `EventArgs` there are several steps necessary to make that work:

Change
https://github.com/mapbox/mapbox-unity-sdk/blob/5a60ead8937d1378945f36208f62a28b513ed427/sdkproject/Assets/Mapbox/Core/mapbox-sdk-cs/Map/Tile.cs#L117

to
```c#
public event EventHandler<TileErrorEventArgs> OnTileError;
```


Change 
https://github.com/mapbox/mapbox-unity-sdk/blob/5a60ead8937d1378945f36208f62a28b513ed427/sdkproject/Assets/Mapbox/Core/mapbox-sdk-cs/Map/Tile.cs#L213-L216

to
```c#
EventHandler<TileErrorEventArgs> handler = OnTileError;
if (handler != null)
{
  handler(this, new TileErrorEventArgs(_id, this.GetType(), null, _exceptions));
}
```

Change
https://github.com/mapbox/mapbox-unity-sdk/blob/5a60ead8937d1378945f36208f62a28b513ed427/sdkproject/Assets/Mapbox/Unity/MeshGeneration/Data/UnityTile.cs#L250-L257

to
```c#
private void OnTileErrorHandler(object sender, TileErrorEventArgs e)
{
  EventHandler<TileErrorEventArgs> handler = OnTileError;
  if (handler != null)
  {
    e.UnityTileInstance = this;
    handler(this, e);
  }
}
```


Change 
https://github.com/mapbox/mapbox-unity-sdk/blob/5a60ead8937d1378945f36208f62a28b513ed427/sdkproject/Assets/Mapbox/Unity/Map/AbstractMapVisualizer.cs#L53

to
```c#
public event EventHandler<TileErrorEventArgs> OnTileError;
```

Change
https://github.com/mapbox/mapbox-unity-sdk/blob/5a60ead8937d1378945f36208f62a28b513ed427/sdkproject/Assets/Mapbox/Unity/Map/AbstractMapVisualizer.cs#L177-L180

to
```c#
EventHandler<TileErrorEventArgs> handler = OnTileError;
if (handler != null)
{
  handler(sender, e);
}
```

Change
https://github.com/mapbox/mapbox-unity-sdk/blob/5a60ead8937d1378945f36208f62a28b513ed427/sdkproject/Assets/Mapbox/Unity/MeshGeneration/Factories/AbstractTileFactory.cs#L44

to
```c#
public event EventHandler<TileErrorEventArgs> OnTileError;
```

Change
https://github.com/mapbox/mapbox-unity-sdk/blob/5a60ead8937d1378945f36208f62a28b513ed427/sdkproject/Assets/Mapbox/Unity/MeshGeneration/Factories/AbstractTileFactory.cs#L66-L72

to
```c#
private void Tile_OnTileErrorEvent(object sender, TileErrorEventArgs e)
{
  EventHandler<TileErrorEventArgs> handler = OnTileError;
  if (handler != null)
  {
    handler(this, e);
  }
}
```

In case I missed anything in the above list here's my branch for comparision:
https://github.com/mapbox/mapbox-unity-sdk/tree/ErrorCallback-bergwerkgis

----




Please replace spaces with tabs as per our [coding style definition](https://github.com/mapbox/mapbox-unity-sdk/blob/develop/CODING-STYLE.md).

Hint: switch on display of Whitespaces in your IDE.

To prevent coding style issues in the future setup your IDE of choice as much as possible to automatically format according to our style automatically: braces on new line, tabs as whitespace, ...

Spaces instead of tabs:

https://github.com/mapbox/mapbox-unity-sdk/blob/5a60ead8937d1378945f36208f62a28b513ed427/sdkproject/Assets/Mapbox/Core/mapbox-sdk-cs/Map/Tile.cs#L114-L116

https://github.com/mapbox/mapbox-unity-sdk/blob/5a60ead8937d1378945f36208f62a28b513ed427/sdkproject/Assets/Mapbox/Unity/Map/AbstractMapVisualizer.cs#L51-L53

https://github.com/mapbox/mapbox-unity-sdk/blob/5a60ead8937d1378945f36208f62a28b513ed427/sdkproject/Assets/Mapbox/Unity/Map/AbstractMapVisualizer.cs#L79-L83

https://github.com/mapbox/mapbox-unity-sdk/blob/5a60ead8937d1378945f36208f62a28b513ed427/sdkproject/Assets/Mapbox/Unity/Map/AbstractMapVisualizer.cs#L92-L96

https://github.com/mapbox/mapbox-unity-sdk/blob/5a60ead8937d1378945f36208f62a28b513ed427/sdkproject/Assets/Mapbox/Unity/Map/AbstractMapVisualizer.cs#L107-L110

https://github.com/mapbox/mapbox-unity-sdk/blob/5a60ead8937d1378945f36208f62a28b513ed427/sdkproject/Assets/Mapbox/Unity/Map/AbstractMapVisualizer.cs#L175-L181

https://github.com/mapbox/mapbox-unity-sdk/blob/5a60ead8937d1378945f36208f62a28b513ed427/sdkproject/Assets/Mapbox/Unity/MeshGeneration/Factories/AbstractTileFactory.cs#L7-L8

Lots of mixed tabs/spaces in
https://github.com/mapbox/mapbox-unity-sdk/blob/5a60ead8937d1378945f36208f62a28b513ed427/sdkproject/Assets/Mapbox/Unity/MeshGeneration/Factories/AbstractTileFactory.cs

apavani:
Those were great points @BergWerkGIS 
I have incorporated the changes as suggested. I have added a RequireComponent (typeof(TileErrorHandler)) to the AbstractMap class to force that script on the map GameObject every time someone adds a component inheriting from that class

BergWerkGIS:
3.
Again getting two error logs per error after attaching the error handler:
image

apavani:
@BergWerkGIS 
How do you spot those indentation problems? 😆

atripathi-mb:
>We still have to decide, if we attach the error handler per default or not?

@apavani @BergWerkGIS I vote for attached by default. If this is optional, a lot of people are going to forget/overlook it. Also, errors handling in general allows for cleaner flow of control.

brnkhy:
`We still have to decide, if we attach the error handler per default or not?`
OK this might be little off topic (and we can move to a ticket if you want) but this talked reminded me this;
https://twitter.com/joegrainger/status/918858834684325888

I have no idea if there's an easy way to do this (other than inheriting all modules from same class, which is a good idea anyway, and handle errors there. that's actually quite simple, I like that).

apavani:
screen shot 2017-11-07 at 1 13 09 pm
I stumbled upon this error when trying to generate tile errors. I set the range tile provider values to 5 , 5, 5, 5 and clicked play. As soon as a few tiles loaded, I stopped playing the scene and immediately following that I saw these errors in VectorLayerVisualizer.

Should we add a separate ticket to resolve this? We just need 
```
if (!tile)
{
  return;
}
```

apavani:
>Again getting two error logs per error after attaching the error handler:

I couldn't reproduce this issue. Did you have two error handlers attached by any chance.
As of the latest commit, `AbstractMap` has a ``
that automatically adds the `TileErrorHandler` Script to the GameObject
Or, did you used add a component that is subscribing to the OnTileError event from the inspector, that is also logging the event along with the default logger?

BergWerkGIS:
@apavani please create a separate ticket with your findings about UnityTile and MissingReferenceException.

---

> How do you spot those indentation problems?

With 
* `Visual Studio Code`
* display of whitespaces turned on, setting: `"editor.renderWhitespace": "all"`
* and `Git Tree Compare` 

as outlined in [my devlog](https://github.com/mapbox/hey/issues/6578):
image

----

> I couldn't reproduce this issue. Did you have two error handlers attached by any chance.
> As of the latest commit, `AbstractMap` has a `[RequireComponent (typeof(TileErrorHandler))]`

Yesterday I had https://github.com/mapbox/mapbox-unity-sdk/commit/c738818ec728c0e326582eba70116a4ef3d6c6b3 checked out which includes `[RequireComponent (typeof(TileErrorHandler))]`.
But I didn't see any errors in the console so I attached errorhandler manually to `Map` of `Globe example`.
Then I got every error twice.

Same today, pulled latest commit https://github.com/mapbox/mapbox-unity-sdk/commit/7464fd2b5fd8e80091f5ab64a4b705a23a1d7e3c and no errors when using as is:

image

After attaching `TileErrorHandler.cs` I get every error twice.
Also notice `List is Empty` on the script. Should it contain the errors?
image

david-rhodes:
screen shot 2017-11-14 at 11 14 07 am

I'm also getting duplicate exceptions. We should find out what layer duplicates are being generated in. If it's at a level lower than this PR, we should cut a new ticket to fix those.

apavani:
@david-rhodes @BergWerkGIS I removed OnTileError event handler form UnityTile and Tile classes. There was some spaghetti of callbacks happening before this. I had to keep it simple such that the Factory concrete implementations will call a virtual method on AbstractTileFactory after checking for the HasError property on the raw tile instance.

Issue #264, opened 2017-09-27
Update DynamicZoomCameraMovement to allow panning with drag


david-rhodes:
As opposed to on mouse up.

Issue #137, opened 2017-06-28
Zoom level change to float type


xwu93:
Do we have the future plan to change the zoom parameter from int to float, give more flexible to show the map? I know the SDK for other platforms have the zoom level as float type.

I looked at the code, however, in someplace, it uses shift operator for two int type, as far as I know, you can't do shifting floating point types in C#, right?

Do you have other way to change the zoom effect more precisely?

Thanks.

brnkhy:
@xwu93 I asked around and tried to understand the effects of this but I'm still not sure if I get it.
So data only changes in integer values like 13-14-15, and as far as I know, tile coverage only changes with those as well. So what's the difference between 15.4 and 15? 
In our javascript maps, 15.4 will look a little more zoomed in but that's all about camera work and nothing about the actual data and/or output. 
This shouldn't be too hard I guess but can you explain a little more why do you need this?

david-rhodes:
@xwu93 Thanks for the suggestion! We'll keep this in mind as we address this issue: https://github.com/mapbox/mapbox-unity-sdk/issues/121.

david-rhodes:
Map components now have a float representation as well.

PR #321, opened 2017-10-25
Dynamiczoom v3 

atripathi-mb:
First pass on Dynamic Zoom with static camera. 
Builds up on Dynamic zoom v2 with the following changes : 

1. Scales unity tiles based on current zoom level. Added an Initialzoom property to UnityTile
(Slippy Map Example now works when zoom level is changed at run time with this change)
2. Scales the Map to fill the viewport. Added and initalzoom property to map to correctly calculate the scaling factor. 
3. Adds a float ZoomRange Property to AbstractMap which drives the maps actual Zoom level. Should address  #137 
4. Adds a PanRange & MapPanned property to AbstractMap to pan the map.
5. Works with BasicMap
Use the QuadTreeLoaderTestScene to test. 

Known issues : 
1. Tiles initialized at higher zoom levels, don't reset their terrain geometry. This results in spiky tiles at lower zoom levels. 
2. MapGenerationBasics example is broken when changing zoom at runtime (works fine is zoom level is fixed) 

Should address the following issues 
#264 
#240


brnkhy:
@atripathi-mb checking this now, a few questions;
1) do you think is it possible to replace int zoom with float zoomRange all together? Do we need the int one?
2) I feel like pan logic doesn't belong to AbstractMap, do you think it would be possible to move that somewhere else? I think you're using that just to calculate the zooming point? maybe we can make that work with a transform or something then use camera (or cam pointer) for reference? something like that would allow other nice stuff as well I guess (like using game character for all that)
3) It loads lots of tile on top of each other when I rotate camera, are you planning to support that as well?

atripathi-mb:
@brnkhy 
>do you think is it possible to replace int zoom with float zoomRange all together? Do we need the int one?

Yes, this is the first step towards replacing the int value with a float one. I did not want to break all the examples, so left it in there. Will clean it up once I have a little better understanding of the code. 

> I feel like pan logic doesn't belong to AbstractMap, do you think it would be possible to move that somewhere else? I think you're using that just to calculate the zooming point? maybe we can make that work with a transform or something then use camera (or cam pointer) for reference? something like that would allow other nice stuff as well I guess (like using game character for all that)

Agreed, only other place I could think of placing it is in the MapTileProvider. Would that be better? My aim was to decouple the camera from the map logic, therefore moved the logic out to the map.

>It loads lots of tile on top of each other when I rotate camera, are you planning to support that as well?

Yes, rotating the camera should work now. I pushed a fix to fix the "tiles on top" issue.

atripathi-mb:
@brnkhy @david-rhodes  
Moved pan related properties into QuadTreeTileProvider from AbstractMap. Let me know if you see any other issues. 
I will start another ticket to fix the map going above the camera issue.

david-rhodes:
@atripathi-mb Not sure what's supposed to happen, but changing `UnityTileSize` at runtime will hang unity. :trollface:

david-rhodes:
@atripathi-mb Can you add a single method call for changing location/zoom at runtime, please?

atripathi-mb:
@brnkhy  
I changed the int zoom value to a float zoomRange value. There is a read only property Zoom which will return the integer zoom level that was requested from the services. 

@david-rhodes 
The UpdateMapProperties method should now take the current zoom level, instead of the delta. Added some checks to handle edge cases as well. This helps de-couple the camera from the map more (QuadTreeCameraMovement.cs now uses this function to pan/zoom) . I see the new imagery loading, let me know if it still is a problem for you.

Shoutout to @david-rhodes for pointing this out, should make the API much simpler :)

BergWerkGIS:
## panning

With mouse:
* doesn't stop when mouse is not moved anymore
* panning diagonally map seems to move along a curve
* at lower zoom levels it feels way too fast
* at higher zoom levels ist feels too slow

With keyboard
* at z0 map jumps whole tile extent
* at z0 it's not possible to position the map in the lower left part of the view


## miscellaneous

* Is it intended that tiles cover the whole viewport?
  If yes, I think formulas need adjustment.
  image
* At higher zoom levels, if the terrain is not flat, terrain moves above camera. This is a known issues and probably out of the scope of this PR.

sdkproject/Assets/Mapbox/Examples/Scripts/DynamicZoomCameraMovement.cs
```diff
@@ -60,7 +64,7 @@ private void LateUpdate()
 			// zoom
 			var y = Input.GetAxis("Mouse ScrollWheel") * _zoomSpeed;
 			//avoid unnecessary translation
-			if (0 != y)
+			if (Mathf.Abs(y) > 0.0f)
```
BergWerkGIS: Just curious, what's the advantage of using `Mathf.Abs`?

sdkproject/Assets/Mapbox/Examples/Scripts/QuadTreeCameraMovement.cs
```diff
@@ -0,0 +1,134 @@
...
+		void Start()
+		{
+			if (null == _referenceCamera)
+			{
+				_referenceCamera = GetComponent<Camera>();
+				if (null == _referenceCamera) { Debug.LogErrorFormat("{0}: reference camera not set", this.GetType().Name); }
+			}
+
+			//put camera facing down. 
+			//_referenceCamera.transform.rotation = new Quaternion(0.7f, 0, 0, 0.7f);
```
BergWerkGIS: For this example it's essential that the camera looks down straight.
I think we should enforce it.
When rotating camera around X-axis the whole effect blows up.

sdkproject/Assets/Mapbox/Examples/Scripts/QuadTreeCameraMovement.cs
```diff
@@ -0,0 +1,134 @@
...
+			_cameraZoomingRangeMinY = (int)(_dynamicZoomMap.UnityTileSize * 1.25f);
+		}
+
+
+		private void LateUpdate()
+		{
+			if (null == _dynamicZoomMap) { return; }
+
+
+			//development short cut: reset center to 0/0 via right click
```
BergWerkGIS: Please remove dead code when it is not needed anymore.

sdkproject/Assets/Mapbox/Unity/Map/AbstractMap.cs
```diff
@@ -9,22 +9,43 @@
...
+		//private int _zoom;
 		public int Zoom
 		{
 			get
 			{
-				return _zoom;
+				return (int)Math.Ceiling(ZoomRange);
 			}
 		}
+        protected int _initialZoom;
```
BergWerkGIS: Please set your IDE to use tabs, not spaces, according to our code style guide:
https://github.com/mapbox/mapbox-unity-sdk/blob/develop/CODING-STYLE.md

sdkproject/Assets/Mapbox/Unity/Map/AbstractMap.cs
```diff
@@ -104,11 +125,14 @@ public void SetCenterLatitudeLongitude(Vector2d centerLatitudeLongitude)
 			_centerLatitudeLongitude = centerLatitudeLongitude;
 		}
 
-		public void SetZoom(int zoom)
+		//public void SetZoom(int zoom)
+		//{
+		//	_zoom = zoom;
+		//}
```
BergWerkGIS: Please remove dead code.

sdkproject/Assets/Mapbox/Unity/Map/AbstractMap.cs
```diff
@@ -117,6 +141,7 @@ void Awake()
 			_fileSouce = MapboxAccess.Instance;
 			_tileProvider.OnTileAdded += TileProvider_OnTileAdded;
 			_tileProvider.OnTileRemoved += TileProvider_OnTileRemoved;
+            _tileProvider.OnTileRepositioned += TileProvider_OnTileRepositioned;
```
BergWerkGIS: Spaces instead of tabs.
Please switch on display of whitespaces in your IDE.

sdkproject/Assets/Mapbox/Unity/Map/AbstractMap.cs
```diff
@@ -127,8 +152,9 @@ void Start()
 		{
 			if (_initializeOnStart)
 			{
-				Initialize(Conversions.StringToLatLon(_latitudeLongitudeString), _zoom);
+				Initialize(Conversions.StringToLatLon(_latitudeLongitudeString), Zoom);
 			}
+            _initialZoom = Zoom;
```
BergWerkGIS: Spaces instead of tabs.

sdkproject/Assets/Mapbox/Unity/Map/AbstractMap.cs
```diff
@@ -180,6 +207,11 @@ void TileProvider_OnTileRemoved(UnwrappedTileId tileId)
 			_mapVisualizer.DisposeTile(tileId);
 		}
 
+        void TileProvider_OnTileRepositioned(UnwrappedTileId tileId)
+        {
+            _mapVisualizer.RepositionTile(tileId);
+        }
+
```
BergWerkGIS: Spaces instead of tabs.

sdkproject/Assets/Mapbox/Unity/Map/AbstractMapVisualizer.cs
```diff
@@ -134,9 +134,9 @@ internal void UpdateState(AbstractTileFactory factory)
 		public virtual UnityTile LoadTile(UnwrappedTileId tileId)
 		{
 			UnityTile unityTile = null;
-
+            //float scaleFactor =
```
BergWerkGIS: Dead code.
Spaces instead of tabs.

sdkproject/Assets/Mapbox/Unity/Map/AbstractMapVisualizer.cs
```diff
@@ -134,9 +134,9 @@ internal void UpdateState(AbstractTileFactory factory)
 		public virtual UnityTile LoadTile(UnwrappedTileId tileId)
 		{
 			UnityTile unityTile = null;
-
+            //float scaleFactor = 
 			if (_inactiveTiles.Count > 0)
-			{
+			{
```
BergWerkGIS: Spaces after `{`.
Please set your IDE to remove those automatically on save.

sdkproject/Assets/Mapbox/Unity/Map/AbstractMapVisualizer.cs
```diff
@@ -177,6 +177,15 @@ public void DisposeTile(UnwrappedTileId tileId)
...
 
+        // TODO: Probably not the best place for this method !!
+        public void RepositionTile(UnwrappedTileId tileId)
+        {
+            UnityTile currentTile;
+            if(ActiveTiles.TryGetValue(tileId ,out currentTile))
+            {
+                PlaceTile(tileId, currentTile, _map);
+            }
+        }
```
BergWerkGIS: Spaces instead of tabs.

sdkproject/Assets/Mapbox/Unity/Map/AbstractTileProvider.cs
```diff
@@ -10,6 +10,7 @@ public abstract class AbstractTileProvider : MonoBehaviour, ITileProvider
 	{
 		public event Action<UnwrappedTileId> OnTileAdded = delegate { };
 		public event Action<UnwrappedTileId> OnTileRemoved = delegate { };
+        public event Action<UnwrappedTileId> OnTileRepositioned = delegate { };
```
BergWerkGIS: Spaces instead of tabs.

sdkproject/Assets/Mapbox/Unity/Map/AbstractTileProvider.cs
```diff
@@ -44,6 +45,17 @@ protected void RemoveTile(UnwrappedTileId tile)
...
+        {
+            if (!_activeTiles.ContainsKey(tile))
+            {
+                //TODO : Only active tiles should be repositioned ?
+                return;
+            }
+
+            OnTileRepositioned(tile);
+        }
+
```
BergWerkGIS: Spaces instead of tabs.

sdkproject/Assets/Mapbox/Unity/Map/BasicMap.cs
```diff
@@ -5,14 +5,16 @@
 	using Mapbox.Map;
 
 	public class BasicMap : AbstractMap
-	{
+	{
```
BergWerkGIS: Spaces instead of tabs.

sdkproject/Assets/Mapbox/Unity/Map/BasicMap.cs
```diff
@@ -5,14 +5,16 @@
...
 
 	public class BasicMap : AbstractMap
-	{
+	{         
 		public override void Initialize(Vector2d latLon, int zoom)
 		{
 			_worldHeightFixed = false;
 			_centerLatitudeLongitude = latLon;
-			_zoom = zoom;
+			//_zoom = zoom;
```
BergWerkGIS: Dead code.

sdkproject/Assets/Mapbox/Unity/Map/DynamicZoomTileProvider.cs
```diff
@@ -92,7 +92,7 @@ void Update()
 			{
 				//already at highest level, don't do anything -> camera free to move closer
 				if (_dynamicZoomMap.Zoom == _dynamicZoomMap.MaxZoom) { return; }
-				_dynamicZoomMap.SetZoom(_dynamicZoomMap.Zoom + 1);
+				_dynamicZoomMap.SetZoomRange(_dynamicZoomMap.Zoom + 1);
```
BergWerkGIS: Just curious: Why the change from `SetZoom` to `SetZoomRange`?
This sets a single value, not a range of values - or am I missing something?

sdkproject/Assets/Mapbox/Unity/Map/IMap.cs
```diff
@@ -1,26 +1,30 @@
 ﻿namespace Mapbox.Unity.Map
```
BergWerkGIS: Whole file: Spaces instead of tabs.

sdkproject/Assets/Mapbox/Unity/MeshGeneration/Data/UnityTile.cs
```diff
@@ -102,6 +102,7 @@ public CanonicalTileId CanonicalTileId
 			}
 		}
 
+        public int InitialZoom { get; internal set; }
```
BergWerkGIS: Spaces instead of tabs.

sdkproject/Assets/Mapbox/Unity/MeshGeneration/Data/UnityTile.cs
```diff
@@ -112,7 +113,9 @@ public CanonicalTileId CanonicalTileId
 		public event Action<UnityTile> OnRasterDataChanged = delegate { };
 		public event Action<UnityTile> OnVectorDataChanged = delegate { };
 
-		internal void Initialize(IMapReadable map, UnwrappedTileId tileId, float scale, Texture2D loadingTexture = null)
+        private bool isInitialized = false;
```
BergWerkGIS: Leading underscore `_` missing on private variable.
Spaces instead of tabs.

sdkproject/Assets/Mapbox/Unity/MeshGeneration/Data/UnityTile.cs
```diff
@@ -121,6 +124,15 @@ internal void Initialize(IMapReadable map, UnwrappedTileId tileId, float scale,
 			_canonicalTileId = tileId.Canonical;
 			_loadingTexture = loadingTexture;
 
+			float scaleFactor = 1.0f; 
+			if (isInitialized == false)
+            {
+				isInitialized = true;
+                InitialZoom = zoom;
+            }
```
BergWerkGIS: Mixed tabs and spaces in the lines above.

david-rhodes:
> Although I don't understand why terrain gets higher on higher zoom levels

Elevation is relative to the zoom level (this affects "world scale"). So, at a low zoom level (2), you're not going to notice any elevation since each tile represents a very large area and the vertex offset is imperceptible. 

The camera is static in this example, but the coordinate system scale is essentially changing at each zoom level. At z14, the pixel/meter ratio is high enough that the relative elevation is approaching the initial camera position height (`y = 135`).

sdkproject/Assets/Mapbox/Examples/Scripts/QuadTreeCameraMovement.cs
atripathi-mb: Camera looking straight down is no longer a requirement. Map is on the ground, and only the part of the map seen in the camera's viewport is drawn.

sdkproject/Assets/Mapbox/Unity/Map/DynamicZoomTileProvider.cs
atripathi-mb: You are correct - Changed Zoom to be a float value, and AbsoluteZoom to return an integer used for server requests.

sdkproject/Assets/Mapbox/Examples/Scripts/DynamicZoomCameraMovement.cs
atripathi-mb: Removes the compiler warning and to avoid precision errors.

atripathi-mb:
@BergWerkGIS 

## panning

With mouse:
>* doesn't stop when mouse is not moved anymore

I am not able to reproduce this, I added a fix for this issue already. 
>* panning diagonally map seems to move along a curve
>* at lower zoom levels it feels way too fast
>* at higher zoom levels ist feels too slow

Changed the factor to damp the movement based on Zoom level. Please try it out and let me know if it feels better. 

With keyboard
>* at z0 map jumps whole tile extent
>* at z0 it's not possible to position the map in the lower left part of the view

Fixed. 

## miscellaneous

>* Is it intended that tiles cover the whole viewport?

Fixed. 

>* At higher zoom levels, if the terrain is not flat, terrain moves above camera. This is a known issue and probably out of the scope of this PR. Although I don't understand why terrain gets higher on higher zoom levels. We'll probably have to revisit the code that extracts height data.

Known issue, not in the scope of this PR. Will address separately.

All formatting issues should also be fixed.

sdkproject/Assets/Mapbox/Examples/Scripts/QuadTreeCameraMovement.cs
BergWerkGIS: Nice! 🎉

sdkproject/Assets/Mapbox/Unity/Map/IMap.cs
```diff
@@ -11,7 +11,9 @@ public interface IMapReadable
 		Vector2d CenterMercator { get; }
 		float WorldRelativeScale { get; }
 		Vector2d CenterLatitudeLongitude { get; }
-		int Zoom { get; }
+		float Zoom { get; }
+		int InitialZoom { get; }
+		int AbsoluteZoom { get; }
```
BergWerkGIS: Please add some comments on what's the difference between these three to make it easier for people learning the SDK.
Also why 2 are `int` and why 1 is `float`.

sdkproject/Assets/Mapbox/Unity/MeshGeneration/Data/UnityTile.cs
BergWerkGIS: still here

sdkproject/Assets/Mapbox/Unity/MeshGeneration/Data/UnityTile.cs
BergWerkGIS: still here

BergWerkGIS:
@atripathi-mb and have you tested all other examples if they still work?

atripathi-mb:
@BergWerkGIS 

> Still getting the missing script error:

Updated the scene, should fix the missing script issue. 

>Panning still feels a bit unrelated to mouse movement.

Tweaked the panning factor some more. I feel this is dependent on the users preference, but we should provide some decent defaults. 

>DynamicZoom example is also affected by this.

DynamicZoom example has not changed other than panning on mouse drag instead of Mouse up. 
 
>Haven't tested this time, but is making it work on device still on your list or should we add to next iteration:
#347

Going to look into adding touch support, if it is more complicated than what I think, I will add it to #347 

> I think we should remove Test from the directory name and from the scene name.
Other examples don't have Test in their names either.

Done, scene & directory name are now called `QuadTreeLoader`

> have you tested all other examples if they still work?
Yes, tested all examples.

One note : Scenes will need to be updated as the zoom parameter has changed from int to float. 
Waiting for all issues to be resolved here before updating scenes.

BergWerkGIS:
> Finally, the pan is fixed. I was using the wrong units to pan, now it should work correctly.

🎉 YES! Feels a lot better now.
Although at lower zoom levels it still feels a bit too slow and at higher zoom levels at bit too fast.
But I'm ok with it the way it is.
I've updated https://github.com/mapbox/mapbox-unity-sdk/issues/347 to revisit during next iteration.

> Added support for touch. Should now work with mouse, keyboard & touch.

Nice! Thanks.
Tested on Android ✅

sdkproject/Assets/Mapbox/Unity/Map/AbstractMapVisualizer.cs
```diff
@@ -134,7 +134,7 @@ internal void UpdateState(AbstractTileFactory factory)
 		public virtual UnityTile LoadTile(UnwrappedTileId tileId)
 		{
 			UnityTile unityTile = null;
-
+			//float scaleFactor =
```
BergWerkGIS: Please remove dead code.

atripathi-mb: Done.

dependencies/Mapbox.IO.Compression-unity/Mapbox.IO.Compression.sln
```diff
@@ -0,0 +1,65 @@
+﻿
```
BergWerkGIS: Please delete this file.
It's a leftover from when we had `Mapbox.IO.Compression-unity` as a submodule.

sdkproject/Assets/Mapbox/Unity/Map/QuadTreeTileProvider.cs
```diff
@@ -0,0 +1,193 @@
...
+			double xDelta = centerLatitudeLongitude.x;
+			double zDelta = centerLatitudeLongitude.y;
+
+			xDelta = xDelta > 0 ? Mathd.Min(xDelta, Mapbox.Utils.Constants.LatitudeMax) : Mathd.Max(xDelta, -Mapbox.Utils.Constants.LatitudeMax);
+			zDelta = zDelta > 0 ? Mathd.Min(zDelta, Mapbox.Utils.Constants.LongitudeMax) : Mathd.Max(zDelta, -Mapbox.Utils.Constants.LongitudeMax);
+
+			_map.SetCenterLatitudeLongitude(new Vector2d(xDelta, zDelta));
+			// Update the center based on current zoom level.
+			var referenceTileRect = Conversions.TileBounds(TileCover.CoordinateToTileId(_map.CenterLatitudeLongitude, _map.AbsoluteZoom));
+			_map.SetCenterMercator(referenceTileRect.Center);
```
BergWerkGIS: Why does the WebMerc center get set to the center of a tile?
Shouldn't this be:
```c#
_map.SetCenterMercator(Conversions.LatLonToMeters(centerLatitudeLongitude));
```

sdkproject/Assets/Mapbox/Examples/Scripts/QuadTreeCameraMovement.cs
```diff
@@ -0,0 +1,172 @@
...
+					ZoomMapUsingTouchOrMouse(zoomFactor);
+					break;
+				default:
+					break;
+			}
+		}
+
+		void ZoomMapUsingTouchOrMouse(float zoomFactor)
+		{
+			_quadTreeTileProvider.UpdateMapProperties(_dynamicZoomMap.CenterLatitudeLongitude, Mathf.Max(0.0f, Mathf.Min(_dynamicZoomMap.Zoom + zoomFactor * 0.25f, 21.0f)));
```
BergWerkGIS: What are those constants?
`0.25` and `21`?

atripathi-mb: 0.25 is the factor by which zoom is incremented/ decremented on mouse scroll.
21 is the highest zoom level for the tileservice

david-rhodes: Should these be serialize fields?

atripathi-mb: Zoom factor should be serializable, highest and lowest zoom levels should probably come from the service.

PR #322, opened 2017-10-25
 Fix for issue #264

atripathi-mb:
Pan on mouse drag instead of mouse up.


david-rhodes:
@atripathi-mb was the intention for scrolling to continue if just holding mouse button down (and not moving the pointer)? Because right now it appears that there is a constant pan, even when my mouse is stationary (and it seems to be accelerating?).

atripathi-mb:
@david-rhodes Good catch, wasn't intentional but I kind of like it ;) Pushed a fix for it though.

david-rhodes:
For clarity, is this fix included in: https://github.com/mapbox/mapbox-unity-sdk/pull/321? If so, do we need this PR?

atripathi-mb:
I can roll this into #321 . Have to port it to QuadTreeCamera anyway. Feel free to discard

david-rhodes:
Done in #321

PR #326, opened 2017-10-26
Add transparency shader for raster tilers

david-rhodes:
This shader allows for custom mapbox studio styles to exclude data types/regions by using the color "black." For example, you could make a label style with white labels and a black background. In Unity, with this shader applied to the material assigned in the `TerrainFactory`, only the labels will be visible.


brnkhy:
@david-rhodes I think there's isn't a public style we can test this easily right? Dark doesn't seem to work, I guess it need to be pitch black?

PR #337, opened 2017-10-31
add FeatureBehaviourModifier

brnkhy:
feature behaviour isn't added to all object by default since last performance merge. and feature behaviour is our only way to keep data on gameobjects so this PR adds a new game object modifier to make that happen again.
what to test;
-open any scene/graph
-choose any non-merged stack and add FeatureBehaviourModifier to GameObject Modifier
-run project, select objects and see FeatureBehaviour script on the inspector, click show properties to check if it's showing

### DIFF
--- a/sdkproject/Assets/Mapbox/Unity/MeshGeneration/Components/FeatureBehaviour.cs
+++ b/sdkproject/Assets/Mapbox/Unity/MeshGeneration/Components/FeatureBehaviour.cs
@@ -7,12 +7,34 @@ namespace Mapbox.Unity.MeshGeneration.Components
 
 	public class FeatureBehaviour : MonoBehaviour
 	{
-		public Transform Transform { get; set; }
+		public VectorEntity VectorEntity;
+		public Transform Transform;
 		public VectorFeatureUnity Data;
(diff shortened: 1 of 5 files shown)


sdkproject/Assets/Mapbox/Unity/MeshGeneration/Components/FeatureBehaviour.cs
```diff
@@ -7,14 +7,36 @@ namespace Mapbox.Unity.MeshGeneration.Components
...
 		public VectorFeatureUnity Data;
 
 		public void ShowDebugData()
 		{
 			DataString = string.Join("\r\n", Data.Properties.Select(x => x.Key + " - " + x.Value.ToString()).ToArray());
 		}
 
+		public void ShowDataPoints()
+		{
+			foreach (var item in VectorEntity.Feature.Points)
```
apavani: Everything looks good to me. The only thing that I want to point out is generally it's best if we use for loop instead of foreach loop. But, the performance impact isn't that much, and considering that this is just used for debugging in this case, we can ignore it.

PR #352, opened 2017-11-06
Enable loading external configurations

ahokinson:
This change allows for the possibility to use a configuration source other than the TextAsset in the Resources folder.

```
public class GlobeMap : AbstractMap
    {
        protected override IEnumerator Start()
        {
            var www = new WWW("https://SomeRemote/MapboxAccess.text");
            yield return www;
            MapboxAccess.ConfigurationJSON = www.text;
            
            yield return base.Start();
```

By default, it will continue to use the TextAsset resource if there is no other configuration JSON specified.

I also fixed a typo.

I have tested these changes in my own project, but because of the classes involved and the execution order changes, the provided SDK examples (and future ones) should be tested.

### DIFF
--- a/sdkproject/Assets/Mapbox/Unity/Map/AbstractMap.cs
+++ b/sdkproject/Assets/Mapbox/Unity/Map/AbstractMap.cs
@@ -1,6 +1,7 @@
 ﻿namespace Mapbox.Unity.Map
 {
 	using System;
+	using System.Collections;
 	using Mapbox.Unity.Utilities;
 	using Utils;
 	using UnityEngine;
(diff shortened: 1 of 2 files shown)


ahokinson:
Unless I'm missing something, this still won't work because the `ConfigureFileSource` and `ConfigureTelemetry` methods need to be run after a configuration is set.

If those methods were run inside of `SetConfiguration`, that would probably work.

david-rhodes:
@ahokinson You are correct. I think we could simply move those from the constructor. You may want to base any changes you make on #430.

atripathi-mb:
@BergWerkGIS Could you take a look at this PR?

BergWerkGIS:
since the external PR branch has been deleted continued at https://github.com/mapbox/mapbox-unity-sdk/pull/552

PR #355, opened 2017-11-07
Performance improvements

brnkhy:
This PR is mainly performance improvements like Count and Vector3/Constant caching.
It also has "MapVisualizerPerformance" monobehaviour, you can throw it on map object and run. it'll create and destroy map bunch of times and average the build time. I think it's better to use this without coroutines enabled.
Then fix some low performance object conversions.

Anyway, I think it would be enough to run demo scenes to test it. It shouldn't change any functionality anyway.


sdkproject/Assets/Mapbox/Unity/MeshGeneration/Modifiers/MeshModifiers/Earcut.cs
```diff
@@ -645,13 +645,22 @@ private static Node insertNode(int i, float x, float y, Node last)
...
-			var dim = data[0][0].Length;
-			var result = new Data() { Dim = dim };
+			var dataCount = data.Count;
+			var dim = 0;
+			for (int i = 0; i < dataCount; i++)
+			{
+				dim += data[i].Count;
+			}
+
+			var result = new Data() { Dim = 2 };
```
atripathi-mb: Why `Dim = 2` ? If we are setting `Dim` to a hardcoded value, what is the use of the for-loop above to compute `dim` ?

sdkproject/Assets/Mapbox/Unity/MeshGeneration/Modifiers/MeshModifiers/Earcut.cs
```diff
@@ -667,13 +676,21 @@ public static Data Flatten(List<List<float[]>> data)
...
-			var dim = 2;
-			var result = new Data() { Dim = dim };
+			var dataCount = data.Count;
+			var dim = 0;
+			for (int i = 0; i < dataCount; i++)
+			{
+				dim += data[i].Count;
+			}
+
+			var result = new Data() { Dim = 2 };
```
atripathi-mb: Why `Dim = 2` ? If we are setting `Dim` to a hardcoded value, what is the use of the for-loop above to compute `dim` ?

sdkproject/Assets/Mapbox/Unity/MeshGeneration/Data/MeshData.cs
```diff
@@ -8,14 +8,14 @@ namespace Mapbox.Unity.MeshGeneration.Data
...
-		public List<List<int>> Triangles { get; set; }
-		public List<List<Vector2>> UV { get; set; }
+		public Vector3 PositionInTile;
+		public List<int> Edges;
+		public Vector2 MercatorCenter;
+		public RectD TileRect;
+		public List<Vector3> Vertices;
+		public List<Vector3> Normals;
+		public List<List<int>> Triangles;
+		public List<List<Vector2>> UV;
```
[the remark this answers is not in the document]
david-rhodes: Yes. Properties are comparable to method calls behind the scenes. With data that is accessed frequently, you could be introducing millions of method calls (and the associated CPU time).

sdkproject/Assets/Mapbox/Unity/Map/AbstractMapVisualizer.cs
```diff
@@ -90,7 +92,7 @@ public void Destroy()
 			// Cleanup gameobjects and clear lists!
 			// This scriptable object may be re-used, but it's gameobjects are likely 
 			// to be destroyed by a scene change, for example. 
-			foreach (var tile in _activeTiles.Values.ToList())
+			foreach (var tile in _activeTiles.Values)
```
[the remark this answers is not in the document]
brnkhy: yep sure, I'll fix that

sdkproject/Assets/Mapbox/Unity/MeshGeneration/Modifiers/MeshModifiers/Earcut.cs
brnkhy: I'm not sure what's up with this. I'm on another branch at the moment but it isn't like this. Maybe mistakenly broken it like that but I have no idea how that could be. 
Anyway dim is always 2 for our case as our raw data is two dimensional anyway so I hardcoded that. I'll switch and check this again as well.

brnkhy: ah now I get it, it's not a mistake, it really is optimization :) 
idea is to calculate the vertex count *2 so set the "result.Vertices" size buuut it's using "dim" field to not initialize yet another int field. 
"dim" field is use to set dimension count in first float overload (original one, not used).
"dim" field isn't even used in vector3 overload.
I'll just rename the second one. and maybe delete that first function all together?

sdkproject/Assets/Mapbox/Unity/Map/AbstractMapVisualizer.cs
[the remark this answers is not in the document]
brnkhy: and the reason this isn't for loop but foreach; that _activeTile property is a dictionary so can't use indexes there. If I used GetEnumarator I'll have to cast object and that feels even worse. I'm not sure what's the better option here.

brnkhy:
@atripathi-mb couldn't fix either of those issues. fixed some smaller stuff, merged dev into it etc. Dim thing is correct, I just renamed the variable. Wasn't sure what to do with that first one (maybe we should keep same stuff in another List to be able to iterate though easier, that's what Unity suggest in similar cases) but we can look into that later as well.

david-rhodes:
screen shot 2017-11-28 at 11 16 17 am

@brnkhy seeing this error in `MeshGeneration.unity` with coroutines enabled in the visualizer and panning the camera to load new tiles.

brnkhy:
@david-rhodes ah nice catch, obvious error though; it's changing the list inside the foreach loop. fixing that now.

Issue #348, opened 2017-11-03
Refactor Events


BergWerkGIS:
Work on technical debt and implement some good practices:

* inherit from `EventArgs`
* replace `Action<> delegate {}` with `EventHandler<>`
  eg `public event Action<TileErrorEventArgs> OnTileError = delegate { };` => 
  `public event EventHandler<TileErrorEventArgs> OnTileError;`
* make them thread safe (we'll never know what the future brings :smirk:)
  instead of
  ```c#
  public event Action<MyEventArgs> TheHandler = delegate { }; 
  if (TheHandler != null) 
  { 
    TheHandler(new MyEventArgs(parameter)); 
  } 
  ```
  use
  ```c#
  public event EventHandler<MyEventArgs> TheHandler;
  EventHandler<MyEventArgs> handler = TheHandler;
  if (handler != null)
  {
    handler(this, new MyEventArgs(parameter));
  }
  ```

References:
*  [`Microsoft Docs: How to: Raise and Consume Events`](https://docs.microsoft.com/en-us/dotnet/standard/events/how-to-raise-and-consume-events)
* [`Microsoft Docs: EventHandler<TEventArgs> Delegate`](https://docs.microsoft.com/en-us/dotnet/api/system.eventhandler-1?view=netframework-4.7.1)

brnkhy:
I don't know why are we going with the Action in the first place. It's not the most common practice but maybe more performant for Unity? @david-rhodes

david-rhodes:
I used to use `EventHandler` with `EventArgs` but `Action` often saves writing code and memory. For example, you can pass objects and value types directly without wrapping inside another object. Every time you use `new` in Unity, a Panda cries 🐼. The fewer of those we have, the better.

Also, with `Action`, you can assign an empty delegate and avoid null checking the event invocation.

BergWerkGIS:
@david-rhodes do I understand correctly that your recommendation is to keep things as they are?

> For example, you can pass objects and value types directly without wrapping inside another object.

From my experience events almost always contain more information than just a single value, eg # of tiles in the queue, tiles left, progress %, errors, ...
And if they don't during initial implementation later refactorings might, forcing changes in a lot of places.
I personally prefer being prepared from the beginning to make things easier down the road (eg multithreading).

But if implementing good practices comes with a performance penalty we might want to steer clear of that for once.

> Also, with `Action`, you can assign an empty delegate and avoid null checking the event invocation. 

[...]

PR #362, opened 2017-11-14
Abstract location provider

david-rhodes:
- Add abstract class wrapper around `ILocationProvider` for most customization with regards to serializing references in the inspector.
- Fixed a bug with iOS locations not updating after initial location update.
- position and rotation with location provider are in local space
- refactored location position and heading into single `Location` object
- All location provider implementations have been updated (and with a bit more functionality for testing/debugging purposes)

I do not currently like the implementation and use of `IsLocationUpdated` and `IsHeadingUpdated` in `Location`. But, I can't think of a good way around this as implemented in `DeviceLocationProvider`. We usually want to update heading as often as possible, and therefore need to create a `Location` object to send out, regardless if our location position has updated. Perhaps we make these values nullable, instead?


BergWerkGIS:
> I do not currently like the implementation and use of `IsLocationUpdated` and `IsHeadingUpdated` in `Location`. But, I can't think of a good way around this as implemented in `DeviceLocationProvider`. We usually want to update heading as often as possible, and therefore need to create a `Location` object to send out, regardless if our location position has updated. Perhaps we make these values nullable, instead?

I think sending `Location` object should only happen when position and/or heading change - to keep number of events as low as possible.
Sending a lot of events and setting properties to `null` to indicate nothing has changed sounds a bit strange to me.

Unless I'm missing something that should be pretty easy:
Cache heading and position like your are doing with `_lastHeadingTimestamp`/`_lastLocationTimestamp` and compare them to new values (maybe even include a small delta to be not too chatty) and only if one or both have changed call `SendLocation()`.
Set `IsHeadingUpdated` and `IsLocationUpdated` accordingly.

----

Is it possible to take orientation (front-/backside) of device into account when evaluating bearing?
Tilting it away/back results in 180° rotation of the map.

Thinking about AR implementations when first looking down and then up a building for instance.

[bearing.mp4.zip](https://github.com/mapbox/mapbox-unity-sdk/files/1474129/bearing.mp4.zip)

----

This https://github.com/mapbox/mapbox-unity-sdk/blob/b0e026f431cbf665ec14678a06ccc0a4d2a6ed09/sdkproject/Assets/Mapbox/Unity/Utilities/Conversions.cs#L61

should be `latLonSplit[1]`

----

So we are going to stick with 
```c#
event Action<Location>
```
instead of inheriting from `EventArgs` and using `EventHandler<>`?
```c#
event EventHandler<LocationUpdatedEventArgs>`
```

https://github.com/mapbox/mapbox-unity-sdk/pull/362/files?diff=split#diff-01a606a52954c00653b1cc722255c2c1

Then please close https://github.com/mapbox/mapbox-unity-sdk/issues/348 as not actionable.

----

Just a personal preference, but I'd prefer to see this code

https://github.com/mapbox/mapbox-unity-sdk/blob/b0e026f431cbf665ec14678a06ccc0a4d2a6ed09/sdkproject/Assets/Mapbox/Examples/LocationProvider/Scripts/MapMatchingExample.cs#L83-L86

formatted like this

image

instead of (also notice the blank in front of `_map`)
image

I think this makes it easier to read and in the (unlikely 😏) case `GeoToWorldPosition` might ever get renamed code still stays aligned.

Same here: https://github.com/mapbox/mapbox-unity-sdk/blob/b0e026f431cbf665ec14678a06ccc0a4d2a6ed09/sdkproject/Assets/Mapbox/Examples/Scripts/PositionWithLocationProvider.cs#L77-L79


-----

Everything spaces instead of tabs except your changes in 
https://github.com/mapbox/mapbox-unity-sdk/blob/abstract-location-provider/sdkproject/Assets/Mapbox/Unity/Location/LocationProviderFactory.cs

If your IDE supports it you could setup it up to auto format on paste/save/edit (and display all whitespaces).
With Visual Studio I have the habit of pressing `Ctrl+KD` (format document) every time before I save.

david-rhodes:
> Cache heading and position like your are doing with _lastHeadingTimestamp/_lastLocationTimestamp and compare them to new values (maybe even include a small delta to be not too chatty) and only if one or both have changed call SendLocation().
Set IsHeadingUpdated and IsLocationUpdated accordingly.

This is a fair point and I considered doing such, but I fear it would not help a whole lot. Heading is practically always changing as a result of the device compass. But, maybe a delta makes sense to prevent jitter anyway.

> `event Action<Location>`

I like this because a struct is "cheaper" than a class. No need to wrap inside `EventArgs`, right?

BergWerkGIS:
> I like this because a struct is "cheaper" than a class. No need to wrap inside `EventArgs`, right? 

I prefer inheriting from `EventArgs` as this also forces you to provide a `sender` which sometimes is handy.
But if performance in Unity is better with `Action` I'm ok with that.

david-rhodes:
> You are still just checking the timestamps and not the actual values.
Is this intended?

Through my testing, it appears that the compass and location data update internally and only change the timestamp when those internal values change. I'm not getting updates from compass if it's sitting on a table, for example.

BergWerkGIS:
> Through my testing, it appears that the compass and location data update internally and only change the timestamp when those internal values change. I'm not getting updates from compass if it's sitting on a table, for example.

Oh wow, interesting.
Would be great to know where this filtering is happening: sensor, device OS, Unity, ....
Definitely have to test on my Droids if they exhibit the same behavior.

sdkproject/Assets/Mapbox/Unity/Location/Location.cs
```diff
@@ -0,0 +1,42 @@
+﻿namespace Mapbox.Unity.Location
+{
+	using Mapbox.Utils;
+	
+	/// <summary>
+	/// Location contains heading, latitude, longitude, accuracy and a timestamp.
+	/// </summary>
+	public struct Location
```
BergWerkGIS: @david-rhodes could you add a `Timestamp` property of type `DateTime` ?
I suppose `Mapbox.Unity.Location` is the object going to be passed over to probe extractor?

david-rhodes: In addition to the current timestamp? Would it be better just to create a static conversion method (unix <--> `DateTime`)

BergWerkGIS: @david-rhodes oops sorry, I totally missed the that there's already a `Timestamp` property 🙈 🙈 🙈 

Conversion methods are already here: https://github.com/mapbox/mapbox-unity-sdk/blob/7f9d203e38c9bbf205c6433bf9de08b8ed9b6a7b/sdkproject/Assets/Mapbox/Core/mapbox-sdk-cs/Utils/UnixTimestampUtils.cs

david-rhodes:
@BergWerkGIS good to merge?

PR #368, opened 2017-11-17
change snap to terrain raycast modifier to take object position and t…

brnkhy:
…ile position into account

change snap to terrain modifier to take object position into account (due to center change options in mod stack)

### DIFF
--- a/sdkproject/Assets/Mapbox/Unity/MeshGeneration/Modifiers/MeshModifiers/SnapTerrainRaycastModifier.cs
+++ b/sdkproject/Assets/Mapbox/Unity/MeshGeneration/Modifiers/MeshModifiers/SnapTerrainRaycastModifier.cs
@@ -28,7 +28,7 @@
 			{
 				for (int i = 0; i < sub.Count; i++)
 				{
-					var h = tile.QueryHeightData((float)((sub[i].x + scaledX / 2) / scaledX), (float)((sub[i].z + scaledY / 2) / scaledY));
+					var h = tile.QueryHeightData((float)((sub[i].x + md.PositionInTile.x + scaledX / 2) / scaledX), (float)((sub[i].z + md.PositionInTile.z + scaledY / 2) / scaledY));
 
 					RaycastHit hit;
 					Vector3 rayCenter =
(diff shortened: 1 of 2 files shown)


sdkproject/Assets/Mapbox/Unity/MeshGeneration/Modifiers/MeshModifiers/SnapTerrainRaycastModifier.cs
```diff
@@ -19,57 +21,31 @@ public override ModifierType Type
...
 
-			if (md.Vertices.Count > 0)
+			foreach (var sub in feature.Points)
 			{
-				for (int i = 0; i < md.Vertices.Count; i++)
+				for (int i = 0; i < sub.Count; i++)
 				{
-					var h = tile.QueryHeightData((float)((md.Vertices[i].x + tile.Rect.Size.x / 2) / tile.Rect.Size.x),
-						(float)((md.Vertices[i].z + tile.Rect.Size.y / 2) / tile.Rect.Size.y));
+					var h = tile.QueryHeightData((float)((sub[i].x + scaledX / 2) / scaledX), (float)((sub[i].z + scaledY / 2) / scaledY));
```
david-rhodes: @brnkhy This does not take into consideration the terrain height multiplier, does it?

david-rhodes:
screen shot 2017-11-17 at 10 47 41 am
Upon closer inspection, it looks like all building inside same tile are set to the same height.

brnkhy:
@david-rhodes I think you had that because you didn't had terrain collider on. Can you check it again please? It took me an hour to realize that as well but all worked properly when I enabled it.

sdkproject/Assets/Mapbox/Unity/MeshGeneration/Modifiers/MeshModifiers/SnapTerrainRaycastModifier.cs
brnkhy: @david-rhodes it should I believe, can't see why not. multiplied height effects queried height anyway right?

david-rhodes:
> I think you had that because you didn't had terrain collider on. Can you check it again please? It took me an hour to realize that as well but all worked properly when I enabled it.

The whole point of your implementation for fallback height was in case the raycast did not hit (no collider for terrain or hitting a tile edge). Therefore, you should test by turning off the terrain collider, right? 😏 At which point, it should use the basic height query.

brnkhy:
@david-rhodes aah you meant that for fallback part, sorry totally missed that and you're right. 👍

brnkhy:
@david-rhodes one problem with that though, it'll be hard to notice it didn't hit anything and using fallback and people will probably think it's raycasting while it actually is not. I'll also try to add an error notification there if it doesn't hit anything.

Issue #377, opened 2017-11-21
Initialize turns off map tiles


HNKMaster:
I'm trying to change the style of my map in-game by modifying the MapUI on the Factory. Everytime I call Initialize() on my map (for example, by pressing Space), my map dissapears. I noticed in hierarchy that every map tile is turned off. When I meddle with the Zoom slider and Initialize again the map (with Space), it works like I want. What's the reason for dissapearing my map?

david-rhodes:
I'm not sure what you mean by map disappearing. But, tiles will be recycled and reloaded with new data when calling `Initialize` on a map object. If you change the style (MapId) of your raster map, a new set of tile requests must be sent to Mapbox servers. This means that we must unload the current tiles and wait for data to be returned from the server to show the new tiles. We use object pooling for this case to save resources and avoid `Instantiate` and `Destroy` calls, which wreak havoc on the GC.

Do you have a code snippet you can share?

HNKMaster:
This is my script for changing syles:

```
void ChangeMapStyle()
{
        factory.MapId = "mapbox://styles/mapbox/" + mapIDs[mapIDCount];
        mapIDCount++;
        if (mapIDCount > 1) mapIDCount = 0;
        map.Initialize(map.CenterLatitudeLongitude, map.Zoom);
        //UpdatePointPosition
}
```

[And this is how my hierarchy looks after using it.](http://oi68.tinypic.com/wukolu.jpg)

david-rhodes:
I just changed `ReloadMap.cs` to the following and everything works as expected:

```cs
using Mapbox.Unity.MeshGeneration.Factories;
using System.Linq;
namespace Mapbox.Examples
{
	using Mapbox.Geocoding;
	using UnityEngine.UI;
	using Mapbox.Unity.Map;
	using UnityEngine;

	public class ReloadMap : MonoBehaviour
	{
		Camera _camera;
		Vector3 _cameraStartPos;
		AbstractMap _map;

		[SerializeField]
		ForwardGeocodeUserInput _forwardGeocoder;

		[SerializeField]
		Slider _zoomSlider;

		[SerializeField]
		Button _button;

		[SerializeField]
		MapImageFactory _factory;

		[SerializeField]
		string[] _mapIds;

		void Awake()
		{
			_camera = Camera.main;
			_cameraStartPos = _camera.transform.position;
			_map = FindObjectOfType<AbstractMap>();
[...]

PR #380, opened 2017-11-21
Fix for tile providers not requesting tiles again 

david-rhodes:
Caused when reinitializing a map object but not updating lat/lon of tile provider.


david-rhodes:
Reference: https://github.com/mapbox/mapbox-unity-sdk/issues/377#issuecomment-346183523

PR #388, opened 2017-11-23
Add ability to read the heightModifier of terrain so it is possible t…

Crashthatch:
…o calculate altitude from the mesh.

---

I had a point on a map collider (from a ray-cast), and needed to know the altitude at that point. But I was displaying terrain with heightModifier set to 3, so point.y was 3x larger that the real-life altitude. Since this value was set in the unity inspector and is private, it's not possible to know what to divide by.


sdkproject/Assets/Mapbox/Unity/MeshGeneration/Factories/TerrainFactory.cs
```diff
@@ -99,6 +99,12 @@ internal override void OnInitialized()
 			_newTriangleList = new List<int>();
 		}
 
+		// Had to add this as we need a way of getting what was set in the Inspector so we can calculate real-life altitude.
+		public float GetHeightModifier()
```
david-rhodes: In this case, it makes sense to make the height modifier public, for changing at runtime.

Issue #384, opened 2017-11-22
Location doesn't update at start (iOS)


mecadiego113:
Hello again mapboxers!

I think I found a bug related with map initialization and location provider.
`InitializeMapWithLocationProvider` uses the device location services to initialize the map with the player position (latitude and longitude).
But, on iPhone, sometimes we don't receive a location update, so the map won't initialize.

Is this a known platform bug or am I doing something wrong?

david-rhodes:
Check in your OS application settings that location is enabled. Also, are you using the latest from `develop`? I've never seen this issue on iPhone, but I have seen slow updates after the initial location is sent (and map loaded). Lastly, try turning Wifi off to see if that helps.

mecadiego113:
Yes, location is always enabled and the app can use it correctly, but it takes a lot to receive the first location update (the map initialization is delayed a lot then).
Disabling WiFi isn't a valid solution for us: our app uses realtime databases (Firebase) to get and set items, positions, stats, etc. and WiFi is a must. 

Is it possible to call a manual update to location services as we do in the last implementation of Editor `EditorLocationProvider`

david-rhodes:
@mecadiego113 Unfortunately, how Unity and iOS handle location services is a bit of a mystery. I've seen some strange behavior with regards to location updates on iOS lately--that's why I suggested you disable Wifi (for testing). My guess is that some combination of Unity/iOS/Device model has some low-level bugs that need to be fixed.

That said, it's also quite possible that my implementation of `DeviceLocationProvider` has some bugs. For example, when the entire sequence was initialized in `Start`, location would not update as often as if it were initialized in `Awake`. Why? I have no idea!

But, feel free to write your own `DeviceLocationProvider` and see if you have better luck.

mecadiego113:
Thank you @david-rhodes!

What I just realized is that if I go to Apple's Maps app (or any other app that uses GPS) and then I go back to my Mapbox app, the location is updated and refreshed, and the map initialized as expected...

I have another question about `AbstractMap`: when the `OnInitialize` event is called? At the very first step of initialization or when it's fully initialized (but not fully downloaded -tiles-).

This is a hack, but I could try to perform a map initialization after a short delay just if it isn't initialized yet.
Should `map.Initialize(new Vector3 (Input.location.latitude, Input.location.longitude), map.AbsolutZoom)` work?

Our next internal release is soon, so I can't research for more iOS location issues now, but I will look into this ASAP and post here what I found.

[...]

PR #389, opened 2017-11-23
Cache the last location update so that it can be read without waiting…

Crashthatch:
… for a new LocationUpdate.

---

Adding a delegate hander to LocationUpdate is fine most of the time, but if you are in a newly created service and immediately want to know the most recent known location, there is no way to get it. This adds a cache of the most recent value to AbstractLocationProvider.

Possibly it would be better to rename this new variable to _currentLocation and refactor subclasses to use that (With my change I think the location gets cached in AbstractLocationProvider AND the subclasses). Also perhaps a getter is better than a public variable?

### DIFF
--- a/sdkproject/Assets/Mapbox/Unity/Location/AbstractLocationProvider.cs
+++ b/sdkproject/Assets/Mapbox/Unity/Location/AbstractLocationProvider.cs
@@ -5,10 +5,12 @@
 
 	public abstract class AbstractLocationProvider : MonoBehaviour, ILocationProvider
 	{
+		public Location _latestLocation;
 		public event Action<Location> OnLocationUpdated = delegate { };
 
 		protected virtual void SendLocation(Location location)
 		{
+			_latestLocation = location;
 			OnLocationUpdated(location);
 		}
 	}
--- a/sdkproject/Assets/Mapbox/Unity/Location/LocationProviderFactory.cs
+++ b/sdkproject/Assets/Mapbox/Unity/Location/LocationProviderFactory.cs
@@ -39,7 +39,7 @@ namespace Mapbox.Unity.Location
 			}
 		}
 
-		ILocationProvider _defaultLocationProvider;
+		AbstractLocationProvider _defaultLocationProvider;
 
 		/// <summary>
 		/// The default location provider. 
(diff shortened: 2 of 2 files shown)


mecadiego113:
May this fix #384? There's a bug using location services on iOS where we don't receive a location update

Crashthatch:
I haven't seen #384 occur in my testing (on Android only) - InitializeMapWithLocationProvider has always managed to get the location immideiatey, but I did have a similar issue where if a LocationUpdate came in to https://github.com/mapbox/mapbox-ar-unity/blob/develop/Assets/Mapbox/Unity/Ar/SimpleAutomaticSynchronizationContextBehaviour.cs before the map had finished initializing (ie. [before the handler was attached](https://github.com/mapbox/mapbox-ar-unity/blob/develop/Assets/Mapbox/Unity/Ar/SimpleAutomaticSynchronizationContextBehaviour.cs#L98) ) it would be missed and so the handler wouldn't run until after the next location update (ie. after walking some distance).

To fix that I modified SimpleAutomaticSynchronizationContextBehaviour to cache any location updates received before the map was initialized, then replay the final one (Call [LocationProvider_OnLocationUpdated](https://github.com/mapbox/mapbox-ar-unity/blob/develop/Assets/Mapbox/Unity/Ar/SimpleAutomaticSynchronizationContextBehaviour.cs#L112) with the cached Location) when the map finishes initializing. There is no PR for this because it's in the AR repo, but I can create one if it would be useful?

With the change this PR introduces, that call to LocationProvider_OnLocationUpdated could use the AbstractLocationProvider._latestLocation rather than maintaining its own cache (neater, less duplication).

It could be a similar issue causing #384. I can't test as I don't have a Mac or iPhone. If so, then this PR won't fix it immediately, but it might make a fix to InitializeMapWithLocationProvider easier.

sdkproject/Assets/Mapbox/Unity/Location/AbstractLocationProvider.cs
```diff
@@ -5,10 +5,12 @@
 
 	public abstract class AbstractLocationProvider : MonoBehaviour, ILocationProvider
 	{
+		public Location _latestLocation;
```
david-rhodes: Prefer getter, and please also add to `ILocationProvider` interface. Naming conventions for public/getter should be capitalized with no preceding `_`.

david-rhodes:
For clarity, `_currentLocation` in subclasses is simply cached to avoid creating a new instance of `Location` every single update. This can probably be refactored in base class by using `protected` or something.

PR #390, opened 2017-11-23
Make it easier to mock GPS positions as though the device was moving …

Crashthatch:
…around a map (eg. provide GPS position of 200m from starting point in x direction).

----

Combine functionality of TransformLocationProvider into EditorLocationProvider so don't have to keep switching which one is being used while testing.

### DIFF
--- a/sdkproject/Assets/Mapbox/Unity/Location/EditorLocationProvider.cs
+++ b/sdkproject/Assets/Mapbox/Unity/Location/EditorLocationProvider.cs
@@ -4,6 +4,7 @@ namespace Mapbox.Unity.Location
 	using Mapbox.Unity.Utilities;
 	using Mapbox.Utils;
 	using UnityEngine;
+	using Mapbox.Unity.Map;
 
 	/// <summary>
 	/// The EditorLocationProvider is responsible for providing mock location and heading data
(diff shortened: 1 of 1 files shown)


sdkproject/Assets/Mapbox/Unity/Location/EditorLocationProvider.cs
```diff
@@ -21,23 +22,38 @@ public class EditorLocationProvider : AbstractEditorLocationProvider
...
 		/// </summary>
 		[SerializeField]
-		[Range(0, 359)]
-		float _heading;
+		Transform _targetTransform;
+
+		[SerializeField]
+		private AbstractMap _map;
+
+		private bool mapInitialized = false;
```
david-rhodes: Where is this ever changed to true?

david-rhodes: Edit: I know what it's for, but it's never set accordingly.

Crashthatch: You're right. I selectively committed hunks to create this PR, so it is set in my version, but not in this commit.
If you're going to create a new PR anyway can I leave it to you to fix?

david-rhodes: No problem!

PR #394, opened 2017-11-24
Implement ProbeExtractor

BergWerkGIS:
Seems to work as expected.

todos left:
* [ ] port remaining methods of https://github.com/mapbox/cheap-ruler
* [ ] tests
  * [ ] cheap ruler
  * [ ] more tests for probe extractor


BergWerkGIS:
Partial port done - all methods needed for GPS filtering should be available now.
Will port remaining ProbeExtractor/CheapRuler methods later on the fly when/if needed.

PR #402, opened 2017-11-27
null Error Handling for GeocodeAttributeSearchWindow

thedrummeraki:
This is just a little `NullReferenceException` handling for `GeocodeAttributeSearchWindow`. I found it very frustrating when the search failed but when the UI was indicating otherwise.

I was thinking of investing the possible cases where the `ForwardGeocodeResponse` response can be `null` on `HandleGeocoderResponse` in order to provide better feedback than "No results found".

### DIFF
--- a/sdkproject/Assets/Mapbox/Unity/Editor/GeocodeAttributeSearchWindow.cs
+++ b/sdkproject/Assets/Mapbox/Unity/Editor/GeocodeAttributeSearchWindow.cs
@@ -129,10 +129,6 @@
 			{
 				_features = res.Features;
 			}
-			else
-			{
-				_features.Clear();
-			}
 			_isSearching = false;
 			this.Repaint();
 
--- a/sdkproject/Assets/Mapbox/Unity/Editor/GeocodeAttributeSearchWindow.cs
+++ b/sdkproject/Assets/Mapbox/Unity/Editor/GeocodeAttributeSearchWindow.cs
@@ -129,8 +129,12 @@
 			{
 				_features = res.Features;
 			}
+			else
+			{
+				_features.Clear();
+			}
 			_isSearching = false;
-			this.Repaint ();
+			this.Repaint();
 
 			//_hasResponse = true;
 			//_coordinate = res.Features[0].Center;
--- a/sdkproject/Assets/Mapbox/Unity/Editor/GeocodeAttributeSearchWindow.cs
+++ b/sdkproject/Assets/Mapbox/Unity/Editor/GeocodeAttributeSearchWindow.cs
@@ -125,9 +125,11 @@
 
 		void HandleGeocoderResponse(ForwardGeocodeResponse res)
 		{
-			_features = res.Features;
+			if (res != null) {
+				_features = res.Features;
+			}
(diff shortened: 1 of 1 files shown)


sdkproject/Assets/Mapbox/Unity/Editor/GeocodeAttributeSearchWindow.cs
```diff
@@ -125,9 +125,12 @@ void HandleUserInput(string searchString)
 
 		void HandleGeocoderResponse(ForwardGeocodeResponse res)
 		{
-			_features = res.Features;
+			if (res != null)
+			{
+				_features = res.Features;
```
david-rhodes: Should you clear `_features` when submitting a new search?

thedrummeraki: Right, good catch! I will test and correct my changes, thank you

thedrummeraki: For some reason, I'm not able to test any cases where the geocoding works (probably because of my network), but I called the `Clear` function on `_features` to make sure it's empty when nothing is found, would that make any sense?

david-rhodes: Ah, so it was actually being reinitialized here: https://github.com/thedrummeraki/mapbox-unity-sdk/blob/07ca43b1b5ca0ccf6fa0ba14f12aee5b02dd623f/sdkproject/Assets/Mapbox/Unity/Editor/GeocodeAttributeSearchWindow.cs#L116. I think you can remove the `else` branch you just added. My bad.

thedrummeraki: Ok cool, didn't realize this myself, I just removed the `else` clause, thanks

david-rhodes:
I don't think I can replicate a failed search (now that I'm testing this PR). Whenever I get a result (with original code) with no features, the UI already displays "no search results." Can you explain what this PR is addressing?

thedrummeraki:
This PR is mostly addressing a simple case when a `null` result from the geocoding is returned. This happens rarely when the internet connection is stable, but I've had experienced issues where a `null` geocode response is returned in the event that the internet becomes unstable. This `null` value not being properly handled, would throw errors.

There are two general cases to consider:

1) One, where no matter what the user searches for (ex: `"Paris France"` or `"hdyieryfg"`), a `ForwardGeocodeResponse` **is** returned with an empty list of features;
2) Another one, where no matter what the user searches for, a `null` `ForwardGeocodeResponse` is returned (what this PR is trying to handle).

The easiest way to replicate a failed search (ie: the second case) is by replicating one of the following sub-cases:

- A missing internet connection
- A internet connection that times out
- A port or URL blocked by the network 

The first sub-case might be the easier to test out. The last two sub-cases were issues experienced within our corporate network.

david-rhodes:
Thank you!

PR #414, opened 2017-11-29
[WIP] Updating LocationProvider example to use Quadtree implementation

atripathi-mb:
First draft -  update LocationProvider to use quadtree tile provider to consolidate examples and create better use cases.


david-rhodes:
Maybe I'm missing something, but the player avatar is not staying in a fixed location, relative to the map as I pan or zoom around.

atripathi-mb:
@david-rhodes @BergWerkGIS , Had to add a new "Quadtree Basic Map" to override the GeoToWorld & WorldToGeo conversions in map space. This is due to the fact that the Quadtree map scales the root to fill up the viewport instead of moving the the camera.

PR #417, opened 2017-11-30
Location provider cached location

david-rhodes:
Replaces #389, #390

### DIFF
--- a/sdkproject/Assets/Mapbox/Unity/Location/EditorLocationProvider.cs
+++ b/sdkproject/Assets/Mapbox/Unity/Location/EditorLocationProvider.cs
@@ -35,8 +35,14 @@ namespace Mapbox.Unity.Location
 
 		protected override void Awake()
 		{
-			base.Awake();
 			_map.OnInitialized += Map_OnInitialized;
+
+			if (_targetTransform == null)
+			{
+				_targetTransform = transform;
+			}
+
+			base.Awake();
 		}
 
 		void Map_OnInitialized()
--- a/sdkproject/Assets/Mapbox/Unity/Location/EditorLocationProvider.cs
+++ b/sdkproject/Assets/Mapbox/Unity/Location/EditorLocationProvider.cs
@@ -4,6 +4,7 @@ namespace Mapbox.Unity.Location
 	using Mapbox.Unity.Utilities;
 	using Mapbox.Utils;
 	using UnityEngine;
+	using Mapbox.Unity.Map;
 
 	/// <summary>
 	/// The EditorLocationProvider is responsible for providing mock location and heading data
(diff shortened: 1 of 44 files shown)


sdkproject/Assets/Mapbox/Unity/Location/EditorLocationProvider.cs
```diff
@@ -21,28 +22,55 @@ public class EditorLocationProvider : AbstractEditorLocationProvider
...
 		{
 			get
 			{
+				if (_mapInitialized)
+				{
+					var startingLatLong = Conversions.StringToLatLon(_latitudeLongitude);
+					var position = Conversions.GeoToWorldPosition(startingLatLong,
+																 _map.CenterMercator,
+																 _map.WorldRelativeScale).ToVector3xz();
+					position += _targetTransform.position;
```
atripathi-mb: Add a null value check for `_transformTransform` , will be nice to be able to use EditorLocationProvider without providing a transform.

sdkproject/Assets/Mapbox/Unity/Location/EditorLocationProvider.cs
```diff
@@ -21,28 +22,55 @@ public class EditorLocationProvider : AbstractEditorLocationProvider
...
+				}
+
 				return Conversions.StringToLatLon(_latitudeLongitude);
 			}
 		}
 
 		protected override void SetLocation()
 		{
-			_currentLocation.Heading = _heading;
+			_currentLocation.Heading = _targetTransform.eulerAngles.y;
```
atripathi-mb: Add a null value check for `_transformTransform` , will be nice to be able to use EditorLocationProvider without providing a transform.

david-rhodes: How about assigning itself if null, instead?

david-rhodes:
> With the addition of target transform to EditorLocationProvider , do we still need TransformLocationProvider?

Great question! I considered this myself and the only thing holding me back from removing it is that some developer may be using it extensively in their project?

sdkproject/Assets/Mapbox/Unity/Location/EditorLocationProvider.cs
atripathi-mb: That works too. I'll make that change.

atripathi-mb:
>Great question! I considered this myself and the only thing holding me back from removing it is that some developer may be using it extensively in their project?

Valid point, let's leave it in for now.

atripathi-mb:
** Sample script to test the conversions**
```
using UnityEngine;
using Mapbox.Utils;
using Mapbox.Unity.Utilities;
using Mapbox.Unity.Map;


public class TestConversions : MonoBehaviour
{

	[SerializeField]
	string _latitudeLongitudeString;

	[SerializeField]
	AbstractMap _map;

	// Use this for initialization
	void Start()
	{


	}

	// Update is called once per frame
	void Update()
	{

		if (Input.GetMouseButtonUp(0))
		{
			Vector2d latLong = Conversions.StringToLatLon(_latitudeLongitudeString);
			var geoToWorld = _map.GeoToWorldPosition(latLong);
			Debug.Log("World : " + geoToWorld.ToString());
			var worldToGeo = _map.WorldToGeoPosition(geoToWorld);
			Debug.Log("Geo : " + worldToGeo.ToString());

		}
	}
}
```

Issue #408, opened 2017-11-29
Geocoding Search not working in Editor. 


atripathi-mb:
To reproduce, click on Search button on the Map Component. Type the name of a city -> no search results. 

In play mode , the search is working. 

Broken since 
https://github.com/mapbox/mapbox-unity-sdk/commit/315be3f61cf507837123283635f05deceaa89266

PR #418, opened 2017-11-30
Fix geocode in editor

BergWerkGIS:
fixes https://github.com/mapbox/mapbox-unity-sdk/issues/408

@atripathi-mb could you please verify that it works again and does not break anything else?


atripathi-mb:
@BergWerkGIS Verified all examples, all good.

BergWerkGIS:
> perhaps we should wrap in #if UNITY_EDITOR the variation? 
sounds good, I'll do that.
BTW: that changes also makes most of the failing tests work again.

PR #431, opened 2017-12-02
Refactor turnstiles to factory for centralization of conditionals

david-rhodes:
new webgl specific telem lib to send user-agent in payload
fallback no longer handles web-gl
use unix utils for timestamp
tileresource queries new factory for events flag


david-rhodes:
@isiyu approves.

PR #448, opened 2017-12-07
Update docs to use new product naming convention

colleenmcginnis:
Updates text to use our new SDK naming conventions.


colleenmcginnis:
Thanks @isiyu!

PR #457, opened 2017-12-07
fix terrain with sidewalls factory naming

brnkhy:


isiyu:
@brnkhy  looks like there are some conflicts here because of updates

PR #467, opened 2017-12-14
Add TagModifier

stevenschobert:
Hey folks! This is a super-small addition, but I needed a way to set a mesh-generated object's tag, so I created this TagModifier (pretty much a copy of the existing LayerModifier).

screen shot 2017-12-14 at 1 55 33 pm

Cheers 🎉!


brnkhy:
This is a nice one @stevenschobert, let's merge this after v1.3 release though.

PR #481, opened 2017-12-31
Fix issue when the CameraBoundsTileProvider's GO is not on world position Vector3.Zero

JustinFincher:
Hi, when I am using the Mapbox SDK I would rather move my camera GameObject than my Map GameObject. Then I found that the Map won't update using the follow setup:

```
Map (GameObject)
- BasicMap (Component)
- CameraBoundsTileProvider (Component)
CameraRig (GameObject)
- MainCamera (GameObject)
-- Camera (Component)
```

If I drag my camera around, the map will populate tiles and fills the rect of my camera. But if I drag my map around the map won't change a thing (leave the rect of the camera blank).  
This is due to the ray-cast result of CameraBoundsTileProvider not being local position but a world position, leading to the wrong `GetGeoPosition` function output as its input is still related to (0,0,0), not a updated position.
If it is not a designated behavior, you can merge this commit. This commit replaced `_ray.GetPoint(_hitDistance)` with `transform.InverseTransformPoint(_ray.GetPoint(_hitDistance))` (assuming that transform of the `CameraBoundsTileProvider` script is the same as the `BasicMap`/`AbstractMap` script)

___
- [ ] Document the PR changes
- [ ] Update the changelog

### DIFF
--- a/sdkproject/Assets/Mapbox/Unity/Map/CameraBoundsTileProvider.cs
+++ b/sdkproject/Assets/Mapbox/Unity/Map/CameraBoundsTileProvider.cs
@@ -57,7 +57,7 @@ namespace Mapbox.Unity.Map
 				_ray = _camera.ViewportPointToRay(_viewportTarget);
 				if (_groundPlane.Raycast(_ray, out _hitDistance))
 				{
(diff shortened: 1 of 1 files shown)


atripathi-mb:
@JustinFincher Thanks for the PR. Will merge this in. 

We added a helper methods to help with WorldToGeo & GeoToWorld position to the map with v1.3.0 based on requests like this.

PR #482, opened 2018-01-03
Adds support to use any color as transparency instead of just black.

atripathi-mb:
Minor fix to add support to use any color for transparency instead of just black. 
___
- [x] Document the PR changes
- [ ] Update the changelog


brnkhy:
looks good to me. Actually quite awesome, I love it! 👍

PR #485, opened 2018-01-04
Removing duplicate definitions of DescriptionAttribute

atripathi-mb:
Replaced the duplicate definitions of DescriptionAtrribute with the correct namespace as per @BergWerkGIS recommendation. 

@BergWerkGIS Could you test this on UWP build please?
___
- [ ] Document the PR changes
- [ ] Update the changelog

### DIFF
--- a/sdkproject/Assets/Mapbox/Core/mapbox-sdk-cs/MapMatching/MapMatchingParameters.cs
+++ b/sdkproject/Assets/Mapbox/Core/mapbox-sdk-cs/MapMatching/MapMatchingParameters.cs
@@ -5,20 +5,10 @@
 //-----------------------------------------------------------------------
 using System;
 using System.ComponentModel;
+using Mapbox.VectorTile.Geometry;
 
 namespace Mapbox.MapMatching
 {
-#if PORTABLE || WINDOWS_UWP
-	[AttributeUsage(AttributeTargets.Field, AllowMultiple = false)]
-	public class DescriptionAttribute : Attribute {
-		private readonly string description;
-		public string Description { get { return description; } }
-		public DescriptionAttribute(string description) {
-			this.description = description;
-		}
-	}
-#endif
-
 	/// <summary>Directions profile id</summary>
 	public enum Profile
 	{
--- a/sdkproject/Assets/Mapbox/Core/mapbox-sdk-cs/Tokens/MapboxTokenApi.cs
+++ b/sdkproject/Assets/Mapbox/Core/mapbox-sdk-cs/Tokens/MapboxTokenApi.cs
@@ -7,17 +7,7 @@ namespace Mapbox.Tokens
 	using Mapbox.Platform;
 	using System;
 	using System.ComponentModel;
-
-#if PORTABLE || WINDOWS_UWP
-	[AttributeUsage(AttributeTargets.Field, AllowMultiple = false)]
-	public class DescriptionAttribute : Attribute {
-		private readonly string description;
-		public string Description { get { return description; } }
-		public DescriptionAttribute(string description) {
-			this.description = description;
-		}
-	}
-#endif
+	using Mapbox.VectorTile.Geometry;
 
 	public enum MapboxTokenStatus
 	{


nosmirck:
I have the following issue:

`Assets\Mapbox\Core\mapbox-sdk-cs\Tokens\MapboxTokenApi.cs(15,4): error CS0104: 'Description' is an ambiguous reference between 'Mapbox.VectorTile.Geometry.DescriptionAttribute' and 'System.ComponentModel.DescriptionAttribute'`

I'm trying to load a simple Map (CitySimulatorMap) into a simple scene for HoloLens.

I tried bypassing this ambiguous definition by forcing the use of `Mapbox.VectorTile.Geometry` like this:

From this `[Description("Point")]` to `[Mapbox.VectorTile.Geometry.Description("Point")]`  everywhere, but I get

`The type or namespace name 'Description' does not exist in the namespace 'Mapbox.VectorTile.Geometry' (are you missing an assembly reference?) [Assembly-CSharp]`

If I force it to be `System.ComponentModel.Description` it let me run, but now the VS project doesn't build and throws a bunch or errors (I don't want to link them here unless needed)

To reproduce this, please create an empty project, configure for HoloLens just like this: https://www.mapbox.com/mapbox-unity-sdk/docs/hololens-development.html

Then build and run, you'll see it builds perfectly, deploys to the HoloLens emulator with no problem.

Finally import the MapBox unity package into the Untiy project and put the CitySimulatorMap Prefab in your scene.

From here the problems start, I guess at least part of it is related to this issue.

Note: I did follow this https://www.mapbox.com/mapbox-unity-sdk/docs/windowsstore-uwp-hololens.html when building the project that unity exported after forcing `System.ComponentModel.Description`

BergWerkGIS:
@nosmirck could you tell us what you are using for `Scripting Backend` and `Scripting Runtime`?
We don't yet support `experimental .Net 4.6`.
Also I noticed a bug in the last few versions of Unity where `WINDOWS_UWP` doesn't get defined properly which might causing what you are seeing.

nosmirck:
@BergWerkGIS Here is more information:

- Scripting Runtime Version: Stable (.NET 3.5 Equivalent)
- Scripting Backend: .NET
- API Compatibility Level: .NET 4.6

I still haven't been able to build :(

BergWerkGIS:
> API Compatibility Level: .NET 4.6

If it's possible with your project you could try switching to `NET 2.0` and see if that works.
Although I'm not positive about that.

> I still haven't been able to build :(

I'm sorry, but this is kind of expected as of now.
As I said above recent versions of Unity seem to have changed how/which `#define`s are set and our code is not able to determine HoloLens/UWP platform correctly anymore.

We will have to work through our dependencies (`mapbox-sdk-cs`,` vector-tile-cs`, `Mapbox.Json`, `Mapbox.IO.Compression`) to account for the change in defines being passed by Unity and that will take some time (no ETA available at the moment).

nosmirck:
> As I said above recent versions of Unity seem to have changed how/which #defines are set and our code is not able to determine HoloLens/UWP platform correctly anymore.

Which version should I use? I don't mind downgrading.

BergWerkGIS:
> Should I report this as a bug, with repro steps and all in a separated issue?

Yes, that would be great. Thank you.

> As I said above recent versions of Unity seem to have changed how/which #defines are set and our code is not able to determine HoloLens/UWP platform correctly anymore.

I was able to hunt down the problem.
Actually it seems it was bug in Unity before and it finally has caught up with the documentation ([Platform dependent compilation](https://docs.unity3d.com/Manual/PlatformDependentCompilation.html)):

> **`UNITY_WSA_10_0`**  #define directive for Universal Windows Platform. Additionally `WINDOWS_UWP` is defined when compiling C# files against .NET Core.

Before `WINDOWS_UWP` was defined even when **not** compiling against .NET Core - and we rely on that being available.

----

Going through all the possible combinations of `Runtime`, `Backend` and `Api Level` I found exactly one way to at least export an UWP project from `Unity@2018.1.0.f2` (haven't tried deploying to HoloLens yet, we'll continue in your new ticket if there are other problems during that step - but building in VS worked):

1. Delete/comment the definition of `DescriptionAttribute` in
```
Assets/Mapbox/Core/Plugins/Mapbox/vector-tile-cs/Geometry/Geometry.cs
```
2. Delete all AR stuff not related to HoloLens - don't know if this is necessary but wanted to avoid any possible conflicts upfront.
image
3. Change your settings to:
image
image
4. `XR Settings -> Virtual Reality Supported -> Windows Mixed Reality`
5. `Build Settings -> Build`
6. open exported solution in VS
7. add sqlite to main project: `Add Reference -> Universal Windows -> Extensions -> SQLite for Windows Universal Platform`
image
8. `Build`

PR #494, opened 2018-01-11
Remove redundant content

morgane:
A lot of this content will exist on [this page](https://www.mapbox.com/unity-sdk/), or it already does!

I'm removing it in an effort to centralize where tutorials, technical narratives, and guides exist. All information about the API, changelog, and known issues will stay here.

If anything else absolutely needs to stay here, let me know!

More detail
- installation instructions already exist at /unity-sdk 
- ditto for access token instructions
- ditto for attribution instructions
- the /unity-sdk will hold general explanations about the factory structure
- guides for each example will move to /unity-sdk
- Windows Store/HoloLens pages should also move to /unity-sdk OR known issues (or be removed altogether: thoughts here?)

___
- [ ] Document the PR changes
- [ ] Update the changelog


morgane:
@BergWerkGIS Thanks for calling that out! I've re-included the pages about HoloLens and added more links to /unity-sdk 👍

documentation/docs/00-getting-started.md
```diff
@@ -2,37 +2,17 @@
...
-
-    **OR**
-
-   - Double-click downloaded `.unityPackage`
+Check out the official [Unity SDK page](https://www.mapbox.com/unity-sdk/) for installation steps.
 
 ## Installation from Git (source)
 
 1. `git clone https://github.com/mapbox/mapbox-unity-sdk.git`
 2. [Read about how the project is configured](https://github.com/mapbox/mapbox-unity-sdk/blob/develop/README.md) and [learn how to contribute](https://github.com/mapbox/mapbox-unity-sdk/blob/develop/CONTRIBUTING.md)
```
atripathi-mb: IMO, there should be instructions to run the `update` script here instead of sending the user to the `Readme`

atripathi-mb:
I am totally for _not_ having duplicate/redundant documentation. This is awesome.

documentation/docs/00-getting-started.md
morgane: What's the `update` script? I'm not sure I've used it before.

By the way, the rationale for removing some of this installation content is because we already have instructions at /unity-sdk. Should we update those instructions to mention this script?

atripathi-mb: This is the script I am talking about `update-mapbox-unity-sdk-core` , instructions are here https://github.com/mapbox/mapbox-unity-sdk/blob/develop/README.md#from-this-repository-advanced-latest-development

This script is required to fetch the dependencies for the SDK. Without running the scripts the SDK will not work. 

The instructions on the /unity-sdk page are about getting the package & released versions of the SDK. If someone is cloning the repo - these instructions would be handy. I don't mind having a duplicate of just `Installation from Git` in both places.

morgane: Ohh got it got it. That's a good idea! I'll update the link.

morgane:
Thank you for the feedback y'all! I'm going to talk to @jfurrow about handling redirects for all the removed pages, to make sure we don't end up with broken links. Once we have a good plan for that, I'll merge this.

PR #504, opened 2018-01-16
Fix MapMatching and Decompression tests

### LINKED ISSUE #393: Revisit tests
Looks like latest refactorings are causing quite some tests (23 out of 47) to fail:

image

### LINKED ISSUE #495: Fix MapMatcher tests once and for all
Every now and then some MapMatcher tests fail because the API returns slightly different values for the same requests:

image

Change tests to check if the wrapper contains any values (`!= default values`, meaning parsing works) instead of checking against certain values.

BergWerkGIS:
@atripathi-mb could you please check if all tests are ✅ for you too?
Especially `Decompress` on OSX:

image

* fixes #393 
  * conditional for OSX
* fixes #495 
  * broader range of values to check against


atripathi-mb:
@BergWerkGIS All tests look good. Getting an error for `RequestDnsError` tests. The string being returned is worded differently. 
image

BergWerkGIS:
Dang, the error message seems to change with every other Unity version and platform.

@atripathi-mb thanks. Could you try again?

PR #506, opened 2018-01-18
Editor layer key fix

### LINKED ISSUE #507: Layer key isn't saved when changed through VectorTileFactory inspector
@BergWerkGIS and I realized today that there's a bug in the vector tile factory editor and if you change the layer visualizer keys through vector tile factory editor, it'll reset when you reopen the Unity.

unity_2018-01-18_14-01-44

After inspecting the issue, I realized we're neither changing nor saving the LayerVisualizer serialized object inside VectorTileFactory inspector. So changes are done on the fly but not saved to file. If you select that layer visualizer and change key using its own inspector, it works as intended though.

This goes back to the early days when I wrote all those inspectors. I didn't understood serialized object/properties and how to get/set them properly so I thought I could get away with not using them in certain situations. It worked generally but this is a little different situation where we're changing an object from another object's inspector and turns out doesn't work here.

I patched the issue but still it's not good/optimal fix and we should look into our inspectors again some time soon. Not a critical issue though as it doesn't effect running application and it's all editor side.

PR: https://github.com/mapbox/mapbox-unity-sdk/pull/506
would be nice if someone can test and verify.

brnkhy:
fixes #507 

___
- [ ] Document the PR changes
- [ ] Update the changelog

### DIFF
--- a/sdkproject/Assets/Mapbox/Unity/Editor/VectorTileFactoryEditor.cs
+++ b/sdkproject/Assets/Mapbox/Unity/Editor/VectorTileFactoryEditor.cs
@@ -26,8 +26,9 @@ namespace Mapbox.Editor
 			serializedObject.Update();
 			GUI.enabled = false;
 			script = EditorGUILayout.ObjectField("Script", script, typeof(MonoScript), false) as MonoScript;
+			EditorGUILayout.PropertyField(serializedObject.FindProperty("State"));
 			GUI.enabled = true;
-
+			EditorGUILayout.Space();
 			EditorGUILayout.BeginHorizontal();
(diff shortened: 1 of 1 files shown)


atripathi-mb:
@brnkhy Looks like the `FindProperty` for `State` is throwing a null exception that @BergWerkGIS posted.

atripathi-mb:
Fix works if I comment out line 29.

brnkhy:
@atripathi-mb @BergWerkGIS I just saw these, I'm sorry. 
Removed that line (iirc it was just for debugging/testing) and tested it still works afterwards

PR #521, opened 2018-01-25
Auto refreshing cache

### LINKED ISSUE #261: Update caching mechanisms to support refresh
Two thoughts:
1. Pass in a `ForceRefresh` bool to use web request, regardless if tile is cached
2. Fetch tile timestamp and compare to timestamp of cached tile--use web request if tile is out of date

Some background: https://github.com/mapbox/mapbox-unity-sdk/issues/251

BergWerkGIS:
fix #261 

Updates ambient caching to not only request tiles when the cache has reached its maximum size but also automatically check in the background if a newer version of already locally cached tiles is available.

* If a tile exists in the local cache it is immediately returned.
* Afterwards a HEAD request is sent to the API to check for a newer version of that tile.
* If yes, then the new tile gets downloaded and put into the local cache
* Then the (new) tile is returned to the calling code a second time

Also adds a new item to the `Mapbox` menu to delete all existing cache files not just the one of the current scene.

*TODO:* 
* [x] expose settings in `Mapbox -> Setup` to toggle that behaviour with `default:OFF`
* [x] keep `MapboxAccess.ClearCache()` for backwards compatibility but mark it as deprecated
* [ ] clean up code/remove log statements
* [x] delete test scene

___
- [ ] Document the PR changes
- [ ] Update the changelog


brnkhy:
@BergWerkGIS I tried the steps you posted on slack and it worked as you said. So I saw tile appear, disappear and reload with new data. I couldn't test it any other way though, so let me know if you want me to try anything else.

PR #564, opened 2018-03-05
Fix Turnstile

BergWerkGIS:
**Description of changes**

More checks if access token is set and more error messages to the console to make that clear.

Tested with:
- [x] no config at all
- [x] config with empty token
- [x] config with invalid token
- [x] config with valid token


isiyu:
tested locally with sending turnstiles with valid and invalid tokens 
Looks good to me. 

@atripathi-mb - do you have time to review this as well?

atripathi-mb:
@isiyu @BergWerkGIS Will review towards EOD today. if that works?

PR #586, opened 2018-03-16
Config window 1.4

jim-martin:
**Changes**

PrefabSceneList now uses references to local sub-assets, so serializing example scenes won't break references to those scenes.

AR example Scenes are not yet captured by Mapbox > SerializeExampleScenes, so moving/renaming those scenes will require additional setup.

**QA checklists**

- [ ] Open Mapbox > Setup
- [ ] See this layout: 
screenshot 2018-03-16 10 10 19

- [ ] Run Mapbox > SerializeExampleScenes
- [ ] Verify the Mapbox > Setup layout remains the same.

**Reviewers**

@atripathi-mb , @BergWerkGIS


atripathi-mb:
Looks good.

PR #595, opened 2018-03-19
Revert "Data vis optimization"

atripathi-mb:
Reverts mapbox/mapbox-unity-sdk#594


atripathi-mb:
Not yet ready to merge into develop.

PR #613, opened 2018-03-26
Fix range height extrusion

### LINKED ISSUE #605: What is the values of "Range Height"
0.1 - 0.5
10 - 100
Two set values are all the same result.

svjz _ s c9 koq_ku gus6

d22 ecj m cf w irlj0tzc

atripathi-mb:
**Related issue**

Closes #605 

**Description of changes**

Range Height Extrusion option was not computing the correct wall height for features. The expected behavior is - the properly value gets clipped to min value is lower or gets clipped to max value if higher than max value, unchanged if in between min & max values ( comparison is done before multiplying with the scale factor)

Adds a Scale Factor to scale the values by this value. 

**QA checklists**

- [x] Update the [changelog](https://github.com/mapbox/mapbox-unity-sdk/blob/develop/documentation/docs/05-changelog.md)
- [x] Update documentation.

**Reviewers**

@apavani @brnkhy

### DIFF
--- a/documentation/docs/05-changelog.md
+++ b/documentation/docs/05-changelog.md
@@ -1,14 +1,20 @@
 ﻿## CHANGELOG
+### v.1.4.0
+*??/??/2018*
+
+#####
+- Fix `Range Property` extrusion option for vector features.
+- Add scale factor for extrusion value derived from feature property.
 
 ### v.1.4.0
 *03/20/2018*
 
 #####
-- Drag and drop prefabs for most common use cases. 
+- Drag and drop prefabs for most common use cases.
 - New Abstract Map UI
-- Style development - colorization and atlas template generator 
+- Style development - colorization and atlas template generator
 - Use texture atlas for building geometries.
-- Merge buildings with unique ids using the experimental 3D buildings tileset. 
+- Merge buildings with unique ids using the experimental 3D buildings tileset.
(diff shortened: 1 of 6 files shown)


apavani:
screen shot 2018-03-30 at 10 05 17 am
screen shot 2018-03-30 at 10 06 27 am

apavani:
On a second try I figured that the min value set on the script causes all the buildings to be raised up in air giving us an illusion that snap terrain is not working properly. I guess what's happening is that if the `height` property value is less than min value set in the script, it extrudes the roof tops above the regular height but fails to extrude the side walls for those buildings

atripathi-mb:
@apavani Fixed the floating buildings issue.Please test again to verify 🙏

apavani:
It works fine when:
1. both max and min values are positive, and max > min
2. both max and min values are positive and max < min
3. max is positive and min is negative

However, it does a negative extrusion when max is negative irrespective of the value of min
screen shot 2018-03-30 at 12 57 28 pm

atripathi-mb:
@apavani Added a warning/error message in UI to avoid user errors, also added a check in code to swap values if min height > max height.

apavani:
Looks good. Is the idea that we allow negative values and do a negative extrusion? If yes, then we need to do a proper negative extrusion where the side walls are extruded below the terrain as well

PR #629, opened 2018-03-29
Fix tile position issue with vector features.

### LINKED ISSUE #615: `Point` Layers are always placed on tile center
Layer visualizers with primitive type `point` always appear at tile center, rather than the feature's location. Feature placement options don't seem to have an effect.

Reproduction steps:
1. Create a new abstract map in a new scene
2. Add a new Vector Layer Visualizer
3. Setup the layer visualizer like this:
screenshot 2018-03-26 17 36 43
*Primitive Type: Point
*Feature Position: First Vertex
4. Play the scene and select any poi_label generated on the map, it should appear at tile center.
5. Exit play mode and adjust the `Feature Position` option, every setting produces the same result.

The feature position options supplied seem focused on polygons, could more clear `Point Position` and `Tile Center` options be provided?

* Unity version: 2017.3
* Mapbox SDK version: 1.4

atripathi-mb:
**Related issue**

Closes #615 

**Description of changes**

Property in the inspector was not being set in the actual modifier stack. 
Changed inspector to not show the `Tile Position` option for grouped features. 

**QA checklists**

- [x] Update documentation.

**Reviewers**

### DIFF
--- a/sdkproject/Assets/Mapbox/Unity/Editor/PropertyDrawers/VectorLayerPropertiesDrawer.cs
+++ b/sdkproject/Assets/Mapbox/Unity/Editor/PropertyDrawers/VectorLayerPropertiesDrawer.cs
@@ -182,17 +182,23 @@
 
 		void DrawModifiers(SerializedProperty property, GUIContent label)
 		{
+			var groupFeaturesProperty = property.FindPropertyRelative("coreOptions").FindPropertyRelative("groupFeatures");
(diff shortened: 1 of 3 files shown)


atripathi-mb:
@jim-martin Mainly for performance gains, grouping = lesser draw calls. So ideally you want to group features if you are not interested in them as individual game objects. Note that right now point features don't work very well with grouping. Also, once we have FeatureCollection merged into the SDK, we can technically have all layers be grouped.

PR #685, opened 2018-04-17
Create new POI placement scene with prefabs

morgane:
Add a new POI placement scene, with Mapbox default prefabs. Screenshot example below.

screen shot 2018-04-17 at 2 58 00 pm


willychang21:
will this update 1.4.2??

atripathi-mb:
@willychang21 Yes, this will be available with 1.4.2 release.

PR #748, opened 2018-05-08
Fix layer renaming in treeview

atripathi-mb:
When displaying two tree views in the UI, renaming of rows in second treeview were not working. This was due to some index mismatch in ImGUI treeview. 
 - Fixed by adding a unique Id to index of rows for the second tree. 
 - Refactored the treeview creations to their separate classes. 
 - Fixes layer selection persistance.

**QA checklists**

- [ ] Update documentation.

**Reviewers**


sdkproject/Assets/Mapbox/Unity/SourceLayers/VectorLayer.cs
```diff
@@ -158,7 +154,7 @@ public void Initialize(LayerProperties properties)
 		public void Initialize()
 		{
 			_vectorTileFactory = ScriptableObject.CreateInstance<VectorTileFactory>();
-			_layerProperty.locationPrefabList = LocationPrefabsLayerProperties.locationPrefabList;
+			_layerProperty.locationPrefabList = PointsOfInterestSublayerList;
```
apavani: This is an unnecessary line.

apavani:
Overall it looks fine.. the only catch of this unique id implementation is that we can have 3000 rows before the code will break... but, i cant think of a case unless someone adds so many layers through code

PR #758, opened 2018-05-10
Ui dropdown fix

atripathi-mb:
Fix for dropdown values getting lost on datasource change.


sdkproject/Assets/Mapbox/Examples/1_DataExplorer/Scripts/TextMeshOutline.cs
```diff
@@ -0,0 +1,91 @@
+﻿namespace Mapbox.Examples
```
jim-martin: Why is this  file included here?

apavani: This renders an outline on the text in the Data Explorer scene

PR #928, opened 2018-07-26
Unity Tile recycling bug fix

brnkhy:
fix a bug in data fetchers where they attached data with recycled new (and wrong) tiles on data callback

**QA checklists**

- [ ] Update documentation.

**Reviewers**

@atripathi-mb


brnkhy:
not that this fix is bad but just a few other options, as food for thought;
- we can stop using unity tile reference and just return data along side tile id. Then let factories to handle what to do with those (detect that tile is recycled and dump the data etc). This is kinda good practice because I feel like it's safe to stop using unity tile references around because of recycling.

- UnityTile actually has a cancel method that cancels all data calls belong to itself. But looks like we're still recieving something after it's cancelled? Maybe Tile class cancel call doesn't work as I/We though? Or maybe we're handling it wrong in data fetchers? We actually don't check if tile is cancelled...

- I tried tile.IsCompleted but couldn't be sure if it's working as intended. We should try replacing TileId checks with IsCompleted as well.

atripathi-mb:
Seeing building in the ocean. 
image

PR #936, opened 2018-07-30
Change TileProviders to update only when Map changes.

atripathi-mb:
**Description of changes**

TileProviders calculate extent on every Update loop. This is wasteful, especially for static maps. 
This PR changes the extent calculation to occur only when the map have changed. 

**QA checklists**

- [ ] Update documentation.

**Reviewers**
@brnkhy @greglemonmapbox


greglemonmapbox:
Additional RangeAroundTransform bug details. If I attempt to use RangeAroundTransform as an initial map option, it does not work correctly. If I then use camera bounds and attach camera movement script to cam, that works correctly. If I then use RangeAroundTransform option again, it does work correctly.

brnkhy:
@atripathi-mb testing this out. Camera Bounds tile provider doesn't seem to remove tiles out of FOV, this isn't intentional right? TileProvider returns all tiles in the "active tiles" list, not sure why so far though.

brnkhy:
@atripathi-mb have you merged other stuff into this branch as well? I see CitySimulator scene settings changed and there seems to be an issue with memory.

looks like a leak but it doesn't show any gc allocation in any functions either so I'm confused.

brnkhy:
@atripathi-mb I'm afraid I have no idea but I'll check sometime. I'm not even sure if it's tile provider directly but probably something related. profiler really isn't helping. I gotta find a way to fix/hack that somehow.

Issue #897, opened 2018-07-03
Request: Add total tile count or module state to AbstractTileProvider


DennyLindberg:
In relation to issue #896 I would like to request either functionality in AbstractTileProvider:

**Add current and total tile count**
This tile count should be the number of tiles currently produced by the tile provider. This would allow us to keep track of the current progress of the tile provider.

**Add module state to the tile provider**
As show in #896 we get some unintended behaviour in the code, which could be solved by keeping track of the state of the tile provider. By adding support for a working state we can circumvent other classes like AbstractMapVisualizer finishing too early.

BergWerkGIS:
As mentioned in https://github.com/mapbox/mapbox-unity-sdk/issues/896#issuecomment-402093003 this is in the works (will need some time till it arrives in a release though).

One thing to be aware of and we are trying to integrate as smoothly as possible: there's a chance that the total tile count and state might be continuously changing eg if the user continuously pans around.

DennyLindberg:
Which I think is perfectly fine. 

We wanted to minimize the amount of full-map updates we are doing on our end in regards to some custom paths that we implemented from an external data source. Even if the tiles are continuously loading from the tile provider we found it to be an improvement to wait for the whole group of tiles in the current frame to load rather than re-generate on each separate addition.

atripathi-mb:
Available in 1.4.5

Issue #896, opened 2018-07-03
AbstractMapVisualizer state is spammed when loading cached tiles


DennyLindberg:
_Note: If this is a bug or support ticket, please provide the following information:_
* **Unity version:** 2017.4.4f1
  * **Scripting Runtime Version:** Stable (.NET 3.5 Equivalent)
  * **Scripting Backend:** Mono
  * **Api Compatibility Level:** .NET 2.0 Subset
* **Mapbox SDK version:** 1.4.3
* **The platform you're building to:** Android
* **A description of what you're trying to do:**

In AbstractMapVisualizer.cs there is an action called OnMapVisualizerStateChanged. This callback behaves differently depending on if tiles are cached or not. Let me do a rundown of the current code execution flow and try to point out the problem.

OnMapVisualizerStateChanged is tied to the State variable, which is updated in relation to the factories as they process' new tiles from the TileProvider. If I step through the chain of events in the code it looks something like the following:

TileProvider.AddTile() -> OnTileAdded() -> AbstractMap.TileProvider_OnTileAdded() -> AbstractMapVisualizer.LoadTile() -> AbstractTileFactory.Register() -> Subclasses of AbstractTileFactory.OnRegistered() -> DataFetcher -> Creates a RasterTile -> Which calls MapboxAccess.Request() to get raster data -> CachingWebFileSource.Request() then either asynchronously gets the data from the server **or returns what it has directly**

Which can be seen with this code snippet
```
// immediately return cached tile
callback(Response.FromCache(cachedItem.Data));
```

Now to the problem. When the TileProvider updates or initializes it will add several tiles through a typical synchronous loop. As each tile is added it will trickle down the long chain above. On a first run of the program the raster data must be fetched from the server, which will cause a delay between when the tile is added and when the factory has finished loading. This will cause an offset between the number of tiles added and number of factories running and thus the state of the visualizer will be stable.

However, if there is an existing cache the state of the visualizer will be changed directly after a tile is added and thus each factory instantly returns.

* **Steps to recreate the bug if appropriate:**

Open ZoomableMap. Add a script component containing the code below. (it will log a message whenever a tile is added or the visualizer state changes)

```
using System.Collections;
using System.Collections.Generic;
using UnityEngine;

using Mapbox.Unity.Map;
using Mapbox.Map;

public class BugReport : MonoBehaviour 
{
	void Awake() 
	{
		AbstractMap map = FindObjectOfType<AbstractMap>();

[...]

PR #955, opened 2018-08-10
Map state and Tile state progress enhancement. 

atripathi-mb:
**Related issue**
Addresses #896  #897 

**Description of changes**

Fixes the deterministic state to Unity tile and MapVisualizer.

**QA checklists**

- [ ] Update documentation.

**Reviewers**

@brnkhy @greglemonmapbox


brnkhy:
found a bug which breaks it in different scenes. It works in Explorer scene but when I added loading screen to city explorer and changed it to use `RangeAroundTileProvider` I noticed it doesn't work. 
(I think) What happens is; tile provider passes a _currentExtent object that's kept and maintained for a long time. 
`AbstractMapVisualizer` line 175 `if (_map.CurrentExtent.Count == _activeTiles.Count)` 
returns false even when all tiles are finished because tile provider clears the `_currentExtent` and left side of that equality will return 0. Admitedly I haven't checked why it works in city explorer scene though. It uses a different type of tile provider but that one also clears that list.
to recreate it;
    - open city explorer scene, drag LoadingScreenCanvas to project and create a prefab
    - open city simulator scene, change tile provider to range around transform, add an empty GO in the scene for that
    - add prefab to scene and run

brnkhy:
@atripathi-mb I'm afraid there's still some issues in this. With the latest changes, it fires `finished` right after `working` when I move the map.
to reproduce;
- open city simulator, switch to range around transform
- change `LoadingPanelController` and add `Debug.Log(s);` in line 32 (also enable showing loading screen again if you want, also helps)
- after initial loading, move transform around. loading canvas won't appear and you'll notice working and finished logged at the same time in log.

PR #1018, opened 2018-09-21
Line feature bindings

brnkhy:
I pushed on a separate branch just in case. line feature properties should be working now.

@atripathi-mb

### DIFF
--- a/sdkproject/Assets/ApiTest.cs
+++ b/sdkproject/Assets/ApiTest.cs
@@ -1,12 +1,18 @@
-﻿using System.Collections;
+﻿using System;
+using System.Collections;
 using System.Collections.Generic;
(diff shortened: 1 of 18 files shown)


atripathi-mb:
@brnkhy 
For POI Layers - Prefab doesn't get updated if the category is changed before changing prefab. 
Steps to reproduce. 
1. Add a POI layer called `loc`
2. Set a prefab in the layer, set another prefab in API test script. 
3. Press play. 
4. Using API test menu - `ChangePOICategory` -> Layer is redrawn
5. Using API test menu - `ChangePOIPrefab` -> see new prefab in UI but the old prefab is still being used. 

Note - workd fine if step 4 is not run before step 5.

PR #1058, opened 2018-10-05
remove custom setting from Terrain datasource options

brnkhy:
@atripathi-mb I pushed this PR for removing custom option from terrain source dropbox but I don't like it at all. I think proper way to do this would be leaving everything same but just removing it from the dropbox but I couldn't do it and removed the enum (and all references to it of course). I'm pushing just in case we don't want to spend time on doing a proper fix and merge this as it's not the worst thing eithr.


brnkhy:
well I guess we can just create another enum, without custom but other entries with same index, then use it for ui?

atripathi-mb:
Closing in favor of a less risky solution

PR #1068, opened 2018-10-08
fix resetmap function in AbstractMap

brnkhy:
fix resetmap function in AbstractMap to unregister all tiles from factories, clear caches for all module and reregister all tiles again

@atripathi-mb

### DIFF
--- a/sdkproject/Assets/Mapbox/Unity/Map/AbstractMap.cs
+++ b/sdkproject/Assets/Mapbox/Unity/Map/AbstractMap.cs
@@ -365,9 +365,14 @@ namespace Mapbox.Unity.Map
 		/// Resets the map.
 		/// Use this method to reset the map to and reset all parameters.
 		/// </summary>
+		[ContextMenu("ResetMap")]
 		public void ResetMap()
 		{
-			Initialize(Conversions.StringToLatLon(_options.locationOptions.latitudeLongitude), (int)_options.locationOptions.zoom);
+			//Initialize(Conversions.StringToLatLon(_options.locationOptions.latitudeLongitude), (int)_options.locationOptions.zoom);
+
+			_mapVisualizer.UnregisterAllTiles();
+			_mapVisualizer.ClearCaches();
+			_mapVisualizer.ReregisterAllTiles();
 		}
 
 		public bool IsAccessTokenValid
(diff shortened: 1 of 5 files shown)


sdkproject/Assets/Mapbox/Unity/Map/AbstractMap.cs
```diff
@@ -365,9 +365,14 @@ public virtual void UpdateMap(Vector2d latLon, float zoom)
 		/// Resets the map.
 		/// Use this method to reset the map to and reset all parameters.
 		/// </summary>
+		[ContextMenu("ResetMap")]
```
atripathi-mb: Is this context menu for testing or do we want to commit it?

sdkproject/Assets/Mapbox/Unity/Map/AbstractMapVisualizer.cs
```diff
@@ -274,6 +274,28 @@ private void Factory_OnTileError(object sender, TileErrorEventArgs e)
 			}
 		}
 
+		public void ReregisterAllTiles()
```
atripathi-mb: Rename method to RegisterAllTiles? Why does it need to be re-register?

sdkproject/Assets/Mapbox/Unity/Map/AbstractMap.cs
brnkhy: I would love to have it in actually. My map times out quite a lot of time because of slow internet and with that button I can just reload everything easily so it's pretty useful.

sdkproject/Assets/Mapbox/Unity/Map/AbstractMapVisualizer.cs
brnkhy: I felt like the name `RegisterAllTiles` implies a given list. While in this method we're reregisterin tiles (that already has been registered to map visualizer before) to factories.

atripathi-mb: @brnkhy The method is doing nothing but registering all tiles, calling it re-register might be misleading. Re-register to me means this function will take care of all logic that is required for registering tiles that are already registered.

PR #1146, opened 2018-11-26
Editor preview shared material fix

greglemonmapbox:
**Related issue**

https://github.com/mapbox/unity/issues/735#issuecomment-440370518

From @brnkhy comment:

`if we set material using material field, editor spams Instantiating material due to calling renderer.material during edit mode. This will leak materials into the scene. You most likely want to use renderer.sharedMaterial instead. errors.` **`if we use sharedMaterial, it creates some problems with tiles for example. They all get the last downloaded texture as they're all written on top of same field.`**

**Description of changes**

UnityTile.Recycle:
- Changed texture assignment target from material to shared material

UnityTile.SetRasterData:
- Raster texture no longer assigned directly to material.texture
- Instead, we create a new material instance using MeshRenderer.sharedMaterial as an input-
- we then assign the raster texture to this new material's mainTexture
- then we assign this material back to MeshRenderer.sharedMaterial

**QA checklists**

- [ ] Update documentation.

**Reviewers**

### DIFF
--- a/sdkproject/Assets/Mapbox/Unity/MeshGeneration/Data/UnityTile.cs
+++ b/sdkproject/Assets/Mapbox/Unity/MeshGeneration/Data/UnityTile.cs
@@ -290,12 +290,29 @@ namespace Mapbox.Unity.MeshGeneration.Data
 					_rasterData.Compress(false);
 				}
 
+				Material material;
 
-				Material material = new Material(MeshRenderer.sharedMaterial);
+				if(MeshRenderer.sharedMaterial != null)
+				{
+					material = new Material(MeshRenderer.sharedMaterial);
+
+					if (Application.isEditor && !Application.isPlaying)
+					{
+						DestroyImmediate(MeshRenderer.sharedMaterial, true);
+					}
+					else
+					{
+						Destroy(MeshRenderer.sharedMaterial);
+					}
+				}
+				else
+				{
+					material = new Material(Shader.Find("Standard"));
+				}
+
 				material.mainTexture = _rasterData;
 				MeshRenderer.sharedMaterial = material;
 
-				//MeshRenderer.material.mainTexture = _rasterData;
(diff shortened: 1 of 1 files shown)


greglemonmapbox:
> @greglemonmapbox just a random though, now that we're creating materials (and we'll probably do the same with Meshes) in editor, they'll probably pile up after a little editing and/or moving around.
> I wonder if Unity editor handles unreferenced objects itself or do we still have to call `unloadunusedassets` occasionally to prevent bloating it?

https://docs.unity3d.com/ScriptReference/Resources.UnloadUnusedAssets.html

`Unloads assets that are not used.

An asset is deemed to be unused if it isn't reached after walking the whole game object hierarchy, including script components. Static variables are also examined.`

Given the above, would these materials qualify as 'unsued' ?

greglemonmapbox:
> @greglemonmapbox just a random though, now that we're creating materials (and we'll probably do the same with Meshes) in editor, they'll probably pile up after a little editing and/or moving around.
> I wonder if Unity editor handles unreferenced objects itself or do we still have to call `unloadunusedassets` occasionally to prevent bloating it?

@brnkhy Can you think of a good way to profile this? Would the memory impact of this (if there is one) show up in profiler?

brnkhy:
@greglemonmapbox I believe that method clears up all Unity objects (mesh, material etc) that are not referenced anymore. In example, (in runtime) if you create lots of meshes and just use one of them, you can see unreferenced meshes in profiler memory section (taking a snapshot and checking objects in the scene iirc). Stuff like meshes or materials aren't garbage collected though hence `Resources.UnloadUnusedAssets` exists.
I'm really not sure how it works in the editor though. And I haven't tried it yet but I also don't think profiler will be much help this time. Another way would be simply checking task manager for editor memory usage while spamming world recreation I guess.

brnkhy (approved):
looks good

PR #1184, opened 2018-12-14
change tile providers to be monobehaviours again

brnkhy:
remove unused vector layer in astronaut game demo scene
change tile providers to be monobehaviours again

globe demo scene should work with these changes now. would be good to test changing tile providers (runtime/editor) in any scene as well.

@jordy-isaac @atripathi-mb @greglemonmapbox


atripathi-mb:
Closing in favor of a different fix

PR #1231, opened 2019-01-08
[WIP] Map preview transform cache

brnkhy:
map transform reseting on preview disabled.
it works when map object is parented by another object as well, so you can have a parent object move and rotate it as you wish and map will be created in that local space. i.e. rotating it 90 degrees and having map on a wall

### DIFF
--- a/sdkproject/Assets/Mapbox/Unity/Editor/StyleSearchWindow.cs
+++ b/sdkproject/Assets/Mapbox/Unity/Editor/StyleSearchWindow.cs
@@ -8,6 +8,7 @@
 	using Mapbox.Unity.Utilities;
 	using Mapbox.Unity.Map;
 	using System.Collections;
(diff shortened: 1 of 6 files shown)


atripathi-mb:
@brnkhy The cached position/scale fix doesn't work with the TileProvider fix in the following scenario
1. Enable preview ->Map Draws
2. Press Play.-> Map not rendered. 

Commented the cached position code and added some more checks for editor preview.

brnkhy:
@atripathi-mb I pushed a fix for that issue but found some follow up problems, will look into them now so this is not ready to merge.

brnkhy:
@atripathi-mb I just checked it and actually (other than a small mistake somewhere) it looks fine. I got confused for a moment about `position will reset when I run the app as well` but the proposed way to do it is parent it with another object anyway so I guess that's fine.

PR #1239, opened 2019-01-11
Sidewall only fix

brnkhy:
fix an issue on wall modifier where switching to side-wall-only mode created walls underground (as it skipped pushing original vertices up)


greglemonmapbox:
This works great for buildings, but if I set road layer to `side only`, extrusion is inverted (underground)

image

PR #1247, opened 2019-01-15
[DO NOT MERGE] - Spawn inside fix

brnkhy:
change spawn inside modifier to use mesh bounds instead of collider bounds.

testing;
- open interactive styled vector map scene
- change it to use range around transform with a small (or 0) range.
- move target objects (destroy/create tiles) and observe parks being generated in new tiles. Tree/Bush objects should all be inside parks.


greglemonmapbox:
@brnkhy Seems like things are still spawning outside of park boundaries here. This is at Buena Vista park in SF, not far from where I live, park boundary in red....

image

You can see trees/bushes intersecting buildings in the red scribble area...

image

PR #1250, opened 2019-01-17
Modifier clear fix

brnkhy:
fix a bug where clear method wasn't called before modifier destruction

### DIFF
--- a/sdkproject/Assets/Mapbox/Unity/Map/AbstractMap.cs
+++ b/sdkproject/Assets/Mapbox/Unity/Map/AbstractMap.cs
@@ -790,6 +790,7 @@ namespace Mapbox.Unity.Map
 			if (_options.extentOptions.extentType != MapExtentType.Custom)
 			{
 				ITileProviderOptions tileProviderOptions = _options.extentOptions.GetTileProviderOptions();
+				string tileProviderName = "TileProvider";
 				// Setup tileprovider based on type.
 				switch (_options.extentOptions.extentType)
 				{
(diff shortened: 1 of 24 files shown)


sdkproject/Assets/Mapbox/Unity/Map/AbstractMap.cs
```diff
@@ -497,15 +497,20 @@ protected virtual void Start()
...
-			}
-
+			DestroyChildObjects();
 			// Setup a visualizer to get a "Starter" map.
 			_mapVisualizer = ScriptableObject.CreateInstance<MapVisualizer>();
 		}
 
+		public void DestroyChildObjects()
+		{
+			UnityEngine.Debug.Log("DestroyChildObjects");
```
atripathi-mb: Remove Debug Message.

sdkproject/Assets/Mapbox/Unity/MeshGeneration/Modifiers/MeshModifiers/ReplaceFeatureCollectionModifier.cs
```diff
@@ -46,18 +46,6 @@ public override void Initialize()
...
-			if (_replaceFeatureModifiers != null && _replaceFeatureModifiers.Count > 0)
-			{
-				foreach (var replaceFeatureModifier in _replaceFeatureModifiers)
-				{
-					if (replaceFeatureModifier != null)
-					{
-						replaceFeatureModifier.Clear();
-					}
-				}
-			}
```
atripathi-mb: Why did we remove this code?

greglemonmapbox: I am going to replace this code and push. As mentioned in the above comment, this code was removed when we were exploring the idea of mods self-clearing themselves. We have not implemented that solution, so this code should be aded back in. If anyone has objections we can remove/roll back.

PR #1286, opened 2019-02-13
Streets v8 support

brnkhy:
add street v8 support 

@greglemonmapbox @atripathi-mb


atripathi-mb:
@brnkhy What are we doing about `StreetsWithBuildingIds` ? It is pointing to v7, should we add another enum called `StreetsV8WithBuildingIds` ?

PR #1308, opened 2019-03-11
Remove node editor

brnkhy:
This PR removes the node editor all together as it's not used and I believe broken anyway. I think it's better to just get rid of it instead of trying to fix it even though it's not in our plans moving further.

@atripathi-mb @jordy-isaac


atripathi-mb:
@brnkhy Do we really need to check in changes to the project settings file?

brnkhy:
@atripathi-mb that was a mistake, reverted that.

PR #1384, opened 2019-06-03
world scale strategy fix

### LINKED ISSUE #1378: Mapbox VR scale is wrong
I'm trying to use Mapbox to create an accurate environment for VR.  For example, if a real skyscraper in NYC is 1000m tall in real life, it should be 1000m tall in VR.  I've tried adjusting the zoom scale and changing World Scale to Custom, but no combination of these will make the buildings appear at their actual height.

In hopes of finding a Mapbox-to-Realworld **ratio**, I gathered some data points.
building heights and ratios

Here's my methodology:

- Zoom = 16.
- Scaling Options = Custom, 100 Unity Tile Size.  (Using World Scale created the unusable ratios in column G.)
- Elevation layer type = Flat Terrain.  (Using terrain with elevation made the buildings harder to measure, and it also changed their heights slightly.)
- In the Unity Editor, focus on the building game object (F key), spawn a cube at its center, then increase the cube's y-scale until it was the same height as the building.  This height is recorded in column H.
- Divide building height in meters by it's height in Unity units.  This is the ratio in column I.

From these 7 data points, I calculated an average ratio of **6.12**.  However, I'm not terribly confident in the accuracy of my method, nor do I have any idea what this number signifies.

Does anyone know what 6.12 refers to?
Does anyone know of a precise zoom level where Mapbox creates buildings to their actual scale?

* Unity version: 2018.3.0f2
* Mapbox SDK version: 2.0.0
* The platform you're building to: PC

brnkhy:
change world scale strategy to not include latitude compensation as it's unnecessary in this mode.

latitude compensation is necessary for fixed size tiles where we forces tiles to a fixed value. World scale ar isn't doing any scaling so should be able to use raw height data without any post-scaling.

Closes #1378

@atripathi-mb @jordy-isaac

### DIFF
--- a/sdkproject/Assets/Mapbox/Unity/Map/Strategies/MapScalingAtWorldScaleStrategy.cs
+++ b/sdkproject/Assets/Mapbox/Unity/Map/Strategies/MapScalingAtWorldScaleStrategy.cs
@@ -8,7 +8,7 @@ namespace Mapbox.Unity.Map.Strategies
 		public void SetUpScaling(AbstractMap map)
 		{
 			var scaleFactor = Mathf.Pow(2, (map.AbsoluteZoom - map.InitialZoom));
-			map.SetWorldRelativeScale(scaleFactor * Mathf.Cos(Mathf.Deg2Rad * (float)map.CenterLatitudeLongitude.x));
+			map.SetWorldRelativeScale(scaleFactor);
 		}
 	}
 }


asmccormick:
Thanks for all of your help!  @atripathi-mb You're right -- we cannot apply the new scale factor across X, Y, and Z equally -- this will solve one axis, but it will distort the others.  We must apply a new scale factor to _either_ the horizontal _or_ the vertical, not both.

Here's my latest solution:
1. use World Scale with 16x zoom -- this will build maps with very accurate horizontal distances.  One Unity unit equals one meter.
2. calculate vertical scale factor based upon latitude:
`float verticalScale = 1 / Mathf.Cos(Mathf.Deg2Rad * (float)AbstractMapScript.CenterLatitudeLongitude.x);
`
3. apply _verticalScale_ to both the terrain and the buildings.

I have a problem in Step 3.  It's easy to change Terrain > Exaggeration Factor:
`        AbstractMapScript.Terrain.SetExaggerationFactor(verticalScale);
`
But I can't find a method to change Features > Buildings > Modeling > Scale Factor.

Does such a method exist?

supergra:
In case someone gets stuck on this, go look at the fix @asmccormick  posted on https://github.com/mapbox/mapbox-unity-sdk/issues/1378 a few days later.

PR #1385, opened 2019-06-04
fix astronaut demo scene shader for mobile platforms

brnkhy:
fix astronaut scene building shader for mobile devices.

error was; 
`duplicate system value semantic definition: input semantic 'SV_POSITION' and input semantic 'VPOS'`

I removed VPOS definition from shader, which was created using amplify shader editor. tested for android build (just created build package, couldn't run it) and windows editor/build.

@atripathi-mb @jordy-isaac


brnkhy:
@atripathi-mb @jordy-isaac this should be safe to merge as well but would be cool if one of you can quickly create an android build and see if it goes through.

PR #1406, opened 2019-06-24
change MapId to TilesetId in all files

brnkhy:
we're changing `mapid` to `tilesetid` on all platforms. this PR changes all instanced of mapid both in UI and code (and comments). I think there's only two references to mapid is left and they are in some comments about javascript object/path equivalent so I didn't touch those. 

@atripathi-mb @jordy-isaac


brnkhy:
@atripathi-mb this should be safe to merge, especially if you checked it as well.

PR #1466, opened 2019-09-19
fix broken traffic scene mapbox pin meshes

brnkhy:
mapbox pin prefabs are broken in traffice demo scene, most likely due to upgrading 2019.1.
Fixed the prefab and instances on scene.

@atripathi-mb


atripathi-mb:
@brnkhy Scenes saved in Unity v2019.x will not work in v2017 & 2018. Closing this PR. 

Fix pushed to v2.1.0 branch.

PR #1534, opened 2020-02-06
BugFix for foreign cultures

blipn:
**Related issue**

No issue related

**Description of changes**

BugFix for foreign cultures :
uses of the dot as the decimal separator instead of a coma for some foreign cultures


brnkhy:
Looks good but string interpolation is C# 6 (or 7?) feature right? I wonder if it worths switching away from good old `String.Format`. I believe we don't use newer features too much in SDK at this moment so might help keep it compatible with old versions maybe.

brnkhy:
merging this into another branch, switching back to string format for backwards compatibility and going with that. thanks a lot @blipn !

PR #1669, opened 2020-08-26
Fix time diff calculation bug

mangolas:
**Description of changes**

Kalman filter calculation used wrong field for time diff


shubhamp41:
Should not be merged

mangolas:
Would you care to explain bit more why not? :)

Issue #1671, opened 2020-09-01
Offline map support on public branch for testing


brnkhy:
Hey everyone!
I have been working on offline map support (https://docs.mapbox.com/help/troubleshooting/mobile-offline/) for a while now and wanted to let you know it's on a public branch for testing at the moment.

You can find it here; https://github.com/mapbox/mapbox-unity-sdk/tree/offlineMaps

And usage goes something like this;
```
 var region = new OfflineRegion(
            MapName,
            _minLatLng,
            _maxLatLng,
            _minZoom,
            _maxZoom,
            elevationTilesetId,
            imageryTilesetId,
            vectorTilesetId);
var response = MapboxAccess.Instance.OfflineManager.CreateOfflineMap(region.Name, region);
```

This will download and cache all tiles in this region in given zoom range. I just pushed it today so there might be bugs and issues of course.
Just to be clear, this doesn't mean you can ship data with your application or anything. This is just a variation of regular ambient cache. Regular ambient cache time out, recycle out etc. Offline maps adds a flag to tiles so they won't time out like regular ambient cache tiles. So it ensures those tiles will be in the cache.

In example, let's say you're building a hiking app where people go off grid frequently to hike. You can let them predownload and cache their hiking area in advance, at their home when they have connectivity for example. SDK will ensure cached tiles will not be removed/recycled/deleted so when they go hiking and don't have connection, their tile data will still be in the cache.

This feature also comes with a lot of changes in Sqlite management so please let me know if you notice any bugs and/or performance issues regarding data in general.

brnkhy:
for anyone trying this out, please let me know if you run into troubles, it's `super beta` so there will be bugs but I want to finish this asap so any fixes to this will be priority for me 🙏

Markovicho:
first note: 

OfflineMapDemoController.cs
remove using directive for:

`using UnityEditor.Rendering.LookDev;`

it seems like it's not used and will throw errors if you are not using the right pipeline:
`Assets\Mapbox\Examples\9_OfflineMap\Scripts\OfflineMapDemoController.cs(7,29): error CS0234: The type or namespace name 'LookDev' does not exist in the namespace 'UnityEditor.Rendering' (are you missing an assembly reference?)`

by the way thx for the afford. This feature is pretty important for a lot of showcases. Looking forward to check this out


Postscript:

[...]

PR #1685, opened 2020-09-30
Offline maps

brnkhy:
Add offline map support to Mapbox Unity SDK
It changes almost everything in Sqlite layer as well so needs to be tested. It'll also won't work with old Sqlite file so file should be recreated using `Clear File Cache` method in Mapbox menu in Unity editor.

announcement ticket; https://github.com/mapbox/mapbox-unity-sdk/issues/1671


useronym:
What is the status of this feature?

PR #1807, opened 2021-08-21
Fixed memory leak

Karthick47v2:
**Unity Web Request UploadHandler** isn't disposed correctly, resulting memory leak. There is no code for signaling that the **UploadHandler** is no longer used and cleaning up the resources it is using. I am working on example scene **"ZoomableMap"** for my project. It happened with Unity 2021.1.5f1, Unity 2019.4.29f1 and mobile devices also, all results memory leak and crashes when start playing (90 % of times).

mb

**Description of changes**

Added code to dispose **UploadHandler**

**Reviewers**

@abhishektrip @brnkhy

### DIFF
--- a/sdkproject/Assets/Mapbox/Unity/Telemetry/TelemetryEditor.cs
+++ b/sdkproject/Assets/Mapbox/Unity/Telemetry/TelemetryEditor.cs
@@ -80,6 +80,8 @@ namespace Mapbox.Unity.Telemetry
 
 			postRequest.downloadHandler = new DownloadHandlerBuffer();
 			postRequest.uploadHandler = new UploadHandlerRaw(bodyRaw);
+			
+			postRequest.uploadHandler.Dispose();
 
 			yield return postRequest.SendWebRequest();
 


artoonie (approved):
This is a low-risk change that is necessary for unit tests with mapbox to cleanly pass. Can we please land it?

m0ose (approved):

brnkhy:
@m0ose @artoonie @Karthick47v2 I don't think we'll accept PRs to this repo at this point but I'll definitely take this into account in our next version work.
I don't think I understand the change though; calling dispose right after initializing it? or is it supposed to be after `postRequest.isDone` check?

also change is only in TelemetryEditor class, which shouldn't be active in builds. How can it cause leaks on mobile devices?

Karthick47v2:
@brnkhy, I actually tried placing the dispose call after `postRequest.isDone`, but the issue persisted. It only resolved when I moved it directly after the initialization.

As for the memory leak, I observed it after my Android build would consistently crash after scrolling through the map for 2-3 minutes. Afterward, I used Unity's memory profiler and found that memory usage was steadily increasing.

brnkhy:
@Karthick47v2 thank you very much for the details!
as I mentioned we won't merge this at this point but I'll keep an eye on similar issues for the new versions.

PR #1913, opened 2023-10-06
bugfix _map.SetCenterLatitudeLongitude and _map.UpdateMap() with CultureInfo

izabala033:
When using _map.SetCenterLatitudeLongitude(), this function saves the current latitude and longitude as string. It uses stringFormat to cast Vector2d x and y to string, but doesn't use InvariantCulture like it does at line 319, inside Initialize().
Decimal separator depends on the culture but InvariantCulture uses . which is what you want. In some cases , is used.
The problem is that it was saving Vector2d(1.2, 3.4) as "1,2, 3,4" instead of "1.2, 3.4. This raises an error when using _map.UpdateMap(), when using Conversion.StringToLatLon(). A split is made with "," as separator, raising an ArgumentException("Wrong number of arguments") because the number of arguments are 4 instead of 2.

### DIFF
--- a/sdkproject/Assets/Mapbox/Unity/Map/AbstractMap.cs
+++ b/sdkproject/Assets/Mapbox/Unity/Map/AbstractMap.cs
@@ -1127,7 +1127,7 @@ namespace Mapbox.Unity.Map
 
 		public virtual void SetCenterLatitudeLongitude(Vector2d centerLatitudeLongitude)
 		{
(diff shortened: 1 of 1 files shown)


theolagendijk:
I came to this same fix when a client run into an issue with my app on an iPhone running with Dutch language settings. 

Perfect bug fix for a nasty problem. I would highly recommend that this fix get's pulled in to the main repository. Small change huge improvement for international audience.

PR #1953, opened 2025-09-22
Feature/layer modifier object

tapycha:
**Related issue**

LayerMaskModifierObject is lacking a drop-down selection in the editor, making it hard to work with. 
LayerMaskModifierObject name is inherently wrong and may confuse users. Its work is to set the layer to a gameobject. Gameobject **layer** is an index in the range [0-31]. **LayerMask** usually refers to a bitmask that is used by physics, culling, etc, for layer filtering, where each bit of int value represents a layer via bit shift.  Consider changing the name to LayerModifierObject for clarity.
**Description of changes**

- Changed LayerMaskModifierObject  to LayerModifierObject 
- Added a custom attribute to display layers in the inspector

If approved, probably need to move the  Drawer folder inside the  Editor folder so there is no need for #if #endif.

**QA checklists**

- [ ] Update documentation.

**Reviewers**


brnkhy:
- I think it would be better to move it inside editor folder as you suggested. following the pattern from other modules, `BaseModule/Editor/` would be fitting.
- Naming is an incredibly difficult thing in general but it's even tricky in this case because `LayerDrawer` can be a lot of things. especially since mapbox vector data we are using also consists of lots of `layers` as well. I don't like starting names with `Unity` prefix either but I think `UnityLayerDrawer` might be a better option (or maybe `GameObjectLayerDrawer`?)
- also I think there's some inconsistency with the drawer names but ending with `propertydrawer` is a better idea. later we can change all other mis-named ones to that as well.

I love the PR, I think UI/UX is probably the weakest part of the SDK at the moment and an improvement here is great!

brnkhy:
@tapycha I made bunch of changes here,
I noticed layer and tag stuff are all about vector feature visualizations so we don't need to have them at base module at all. so I moved attributes and drawers to the vector module. it feels better to touch least number of modules I guess. if at any point we need them in base, we can move back.
oh... sure we can add tag&layer options to tile gameobject (TileCreatorBehaviour) and to be able to do that those should be in the base....
well that's another PR for the future :)

PR #1954, opened 2025-09-23
Remove unnecessary delegates, combine/remove from UnityTileImageConta…

tapycha:
…iner

**Related issue**

StaticApiLayerModule, when loading a tile, performs LoadTempTile, then LoadInstant. Both of them call unityTile.ImageContainer.SetImageData, which, upon receiving RasterData, swaps OnDispose delegate. (Removing from old RasterData and adding to a new RasterData). Adding and removing delegates creates allocation, as it creates a new delegate after such an operation. This escalates when loading a lot of tiles at once.
Here is a screenshots from profiler on iPhone 13 Pro Max loading a map at Zoom 18 with TransformBasedTileProviderBehaviour set to 5 in all directions.
Current implementation creates 42KB of allocations and taking almost 11ms of cpu time to proceed
Screenshot 2025-09-23 at 10 52 41
Changing the Combine/Remove delegate to a direct pass of the delegate resolves the issue, but it can cause issues if someone tries to add to these delegates, as that add can be lost as tiles load/unload. A potential solution if this case needs to be handled is to replace the delegate with an array/list and pass it.
Here is a screenshot of the performance of direct pass for the delegate on the same device and map settings.
Screenshot 2025-09-23 at 11 38 45
I'm not sure if RasterData could be used by multiple UnityTileImageContainers, as in that case, this approach is bad.

**Description of changes**

- MapboxTileData removed the default delegate and replaced OnDispose call with a null propagating invoke.
- UnityTileImageContainer removed the default delegate and replaced += and -= with direct pass
- UnityMapTile replaced += with direct pass delegate(probably doesn't change that much)

**QA checklists**

- [ ] Update documentation.

**Reviewers**

### DIFF
--- a/Runtime/Mapbox/BaseModule/Data/DataFetchers/MapboxTileData.cs
+++ b/Runtime/Mapbox/BaseModule/Data/DataFetchers/MapboxTileData.cs
@@ -13,12 +13,17 @@ namespace Mapbox.BaseModule.Data.DataFetchers
         public DateTime? ExpirationDate;
         public bool HasError = false;
         [HideInInspector] public byte[] Data;
+        
+        private Action _onDispose;
 
         public virtual void Dispose()
         {
-            OnDispose?.Invoke();
+            _onDispose?.Invoke();
         }
 
-        public Action OnDispose;
+        internal void SetDisposeCallback(Action callback)
+        {
+            _onDispose = callback;
+        }
     }
 }
--- a/Runtime/Mapbox/BaseModule/Data/DataFetchers/MapboxTileData.cs
+++ b/Runtime/Mapbox/BaseModule/Data/DataFetchers/MapboxTileData.cs
@@ -16,9 +16,9 @@ namespace Mapbox.BaseModule.Data.DataFetchers
 
         public virtual void Dispose()
         {
-            OnDispose();
+            OnDispose?.Invoke();
         }
-        
-        public Action OnDispose = () => { };
+
+        public Action OnDispose;
     }
 }
--- a/Runtime/Mapbox/BaseModule/Unity/UnityMapTile.cs
+++ b/Runtime/Mapbox/BaseModule/Unity/UnityMapTile.cs
@@ -61,10 +61,7 @@ namespace Mapbox.BaseModule.Unity
 
 		public void Awake()
 		{
-			ImageContainer = new UnityTileImageContainer(this)
-			{
-				OnDispose = DataDisposed
-			};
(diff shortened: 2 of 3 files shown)


brnkhy:
wow 10ms, no idea how I missed that.
what do you think about
1) passing it as parameter in constructor
   - no one else can register it
2) calling `_unityMapTile.DataDisposed` directly as we already have reference down in the container 
   - would require dataDisposed being public and calling parent' method is ugly

tapycha:
Passing down as a parameter in the constructor would be a good solution for ```UnityTileImageContainer```, but this can't be done for ```RasterData```, as ```UnityTileImageContainer``` is not creating them, just wrapping around. Maybe for ```RasterData``` swapping ```OnDispose``` to ```internal``` instead of ```public``` would work, but I need to test.
Direct call to ```_unityMapTile.DataDisposed``` would be ugly and would probably create cyclical references. I like the solution with delegates more.

tapycha:
There is probably less impact than 10ms, as this build was a debug build with deep profiling on, but the issue is still there nonetheless.

brnkhy:
`but this can't be done for RasterData`
ah yes I forgot passing to container is only half the problem.
as I described in other PR, whole thing is about invalidating cached tile objects. maybe we can find another way to save us from all these troubles.
but suggested solution can be a temporary fix for now. sure there's a trap where we will forget about this and register to same event later breaking everything down with no easy way to understand what's wrong...
maybe we can make it a method, like `SetDisposeCallback` which would reduce the chances of registering to that method by mistake?
so instead of `ImageData.OnDispose = OnDispose;` it'll be `ImageData.SetDisposeCallback(OnDispose);`

tapycha:
- ```MapboxTileData``` ```OnDispose``` is now ```private``` and added ```SetDisposeCallback``` as ```internal```
- ```UnityTileImageContainer``` moved ```OnDispose``` into constructor and set it to ```private```
- ```UnityTileImageContainer```  changed ```_mainTexFieldNameID```, ```_mainTexStFieldNameID```, ```_mainTextureChangeTimeFieldNameID``` to const and added equivalent int id for shader as a look up. This should speed up a bit ```OnImageryUpdated``` when updating material (probably need to cache this in ```UnityTileTerrainContainer``` too at some point)
-  ```UnityTileImageContainer```  changed ```State```, ```ImageData```,  set accessor to ```private``` for better encapsulation

brnkhy:
@tapycha I think I like this until we find a better solution to invalidate tiles. I'll play with this further and let you know.

tapycha:
Glad to help.

PR #1979, opened 2026-02-26
Map matching api

brnkhy:
fix mapmatching api, add tests

### DIFF
--- a/Documentation~/UsingDirectionsApi.md
+++ b/Documentation~/UsingDirectionsApi.md
@@ -1,0 +1,113 @@
+
+# Using the Mapbox Directions API
+
+The Mapbox Directions API allows you to calculate routes between locations. You can get turn-by-turn directions, route geometry, distance, and duration for different routing profiles (driving, walking, cycling).
+
+For complete API reference and details, see the [official Mapbox Directions API documentation](https://docs.mapbox.com/api/navigation/directions/).
+
+## Basic Setup
+
+To use the Directions API, you need to:
+
+1. Reference the `MapBehaviourCore` component to access the map's file source.
+2. Create a `MapboxDirectionsApi` instance with the file source.
+3. Build a `DirectionResource` with your origin and destination coordinates.
+4. Query the API and handle the response in a callback.
+
+Here's a simple example:
+
+```csharp
+using UnityEngine;
+using Mapbox.Unity.Map;
+using Mapbox.DirectionsApi;
+using Mapbox.DirectionsApi.Response;
+using Mapbox.Utils;
+
+public class DirectionsExample : MonoBehaviour
+{
+    public MapBehaviourCore MapCore;
+    private MapboxDirectionsApi _directionsApi;
+
+    void Start()
+    {
+        MapCore.Initialized += map =>
+        {
+            _directionsApi = new MapboxDirectionsApi(map.MapService.FileSource);
+            GetRoute();
+        };
+    }
+
+    void GetRoute()
+    {
+        // Define two coordinates: San Francisco to Los Angeles
+        var origin = new LatitudeLongitude(37.7749, -122.4194);
+        var destination = new LatitudeLongitude(34.0522, -118.2437);
+
+        // Create a directions request with driving profile
+        var directionResource = new DirectionResource(
+            new[] { origin, destination },
+            RoutingProfile.Driving
+        );
+
+        // Query the API
+        _directionsApi.Query(directionResource, response =>
+        {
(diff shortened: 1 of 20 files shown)


astojilj (commented):
## PR Review: Map matching api

**Author:** @brnkhy | **Branch:** `mapMatchingApi` → `develop` | **+1445 / -54** across 19 files

---

### Summary

This PR fixes the completely broken MapMatching API and adds documentation + example scripts for three APIs:

**Core fix:** `MapMatchingResource.GetUrl()` was returning `null` with all logic commented out. This PR:
1. Introduces `[MapboxRequestString]` attribute to replace `[Description]` for API URL parameter serialization
2. Adds `MapboxTypeDescription()` extension method to read the new attribute
3. Re-enables `GetUrl()` with `.MapboxTypeDescription()` replacing `.Description()`
4. Assigns proper power-of-2 flag values to `Annotations` enum (1, 2, 4, 8)
5. Re-enables the `getUrlQueryFromAnnotations` bitwise iteration

**New content:** Documentation (`UsingDirectionsApi.md`, `UsingGeocodingApi.md`, `UsingMapMatchingApi.md`) and example MonoBehaviour scripts for Directions, MapMatching, Forward Geocoding, and Reverse Geocoding.

---

### Issues

#### 1. ⚠️ `ReverseGeocodeExample.cs` is in the global namespace

All other example scripts are properly namespaced (`Mapbox.DirectionsApi.Example`, `Mapbox.GeocodingApi.Example`), but `ReverseGeocodeExample` is declared at global scope. Should be in `Mapbox.GeocodingApi.Example` to match `ForwardGeocodeExample`.

#### 2. ⚠️ `Annotations` flags enum has no `None = 0` value

```csharp
[System.Flags]
public enum Annotations
{
    [MapboxRequestString("duration")]
    Duration = 1,
    ...
}
```

`[Flags]` enums should conventionally include `None = 0`. Without it, `default(Annotations)` is 0, which has no named member and `Enum.GetName` will return `null` — the `MapboxTypeDescription()` method handles this correctly (falls back to `ToString()`), but it's still a design gap. If someone checks `if (Annotations.HasValue)` with a nullable default, they'd get `0` which is "no flags set" but not explicitly named.

#### 3. ⚠️ Possibly unused `using` in `MapMatchingResource.cs`

```csharp
+using Mapbox.VectorTile.ExtensionMethods;
```

This was where the old `.Description()` extension lived. Since all calls are now `.MapboxTypeDescription()` (from `Mapbox.BaseModule.Utilities`), this import may be unused. Verify and remove if so.

#### 4. ⚠️ Example scripts in Runtime, not Samples~

The example scripts (`DirectionsExample.cs`, `MapMatchingExample.cs`, `ForwardGeocodeExample.cs`, `ReverseGeocodeExample.cs`) are in `Runtime/Mapbox/*/Example/` which means they compile into every project using this package. Unity convention for optional example code is `Samples~` (excluded from compilation by default). This adds 4 MonoBehaviours to the project that most users won't need.

#### 5. 💅 Doc comment says "UrlStringAttribute" in `EnumExtensions.cs`

```csharp
/// Looks for the UrlStringAttribute on the enum value.
```

and

```csharp
// Try to get the UrlStringAttribute
```

Should reference `MapboxRequestStringAttribute` instead of `UrlStringAttribute`.

#### 6. 💅 Reflection on every call in `MapboxTypeDescription()`

The method does `type.GetField(name)` + `Attribute.GetCustomAttribute()` on every invocation. For URL generation this is fine, but if it ever gets used in a hot path, it would need a static cache. Worth a comment noting this trade-off.

#### 7. 💅 `ReverseGeocodeExample.GetAddressFromWorldPosition` has inverted condition

```csharp
if (MapCore == null || MapCore.InitializationStatus <= InitializationStatus.Initialized)
```

This reads like it should be `< InitializationStatus.Initialized` or `!= InitializationStatus.Initialized` — "less than or equal to Initialized" would exclude the valid initialized state. Unless `InitializationStatus` values are ordered differently than expected.

---

### What looks good

- The `[MapboxRequestString]` / `MapboxTypeDescription()` approach cleanly separates API serialization from display concerns — good architectural fix
- The `Annotations` enum now has proper power-of-2 values for bitwise flag operations
- All `.Description()` calls consistently replaced with `.MapboxTypeDescription()`
- The documentation is comprehensive with working code snippets, optional parameters, and response data descriptions
- Example scripts are well-structured with proper null checks, error handling, `[Header]`/`[Tooltip]` attributes, and `OnDestroy` cleanup
- The MapMatching comparison table in the docs is helpful

astojilj (approved):

PR #1981, opened 2026-03-09
Location module changes

brnkhy:
location modules changes; 
integrating mapbox common location system as an option (unity location is still default)
fix device location permissions


astojilj:
Review feedback provided in offline sync

PR #1989, opened 2026-05-18
v3.1.0 merge

brnkhy:
### v3.1.0

Minor-version bump per semver: this release contains source-breaking API changes (listed below) alongside substantial feature work in camera/input, tile LOD, and terrain rendering. The 3.0.7 designation was previously used in development; published 3.0.7 builds, if any, are superseded by 3.1.0.

#### New dependencies
- Added `com.unity.burst@1.8.12`. The terrain-RGB elevation decoder and the collider vertex-fill jobs are now Burst-compiled. First-time domain reload pays a one-shot AOT compile cost; runtime decoding is significantly faster.
- `com.unity.inputsystem` is recognized as a **soft dependency**. Projects without it continue using `UnityEngine.Input`; projects with it automatically use the new Input System via a `MAPBOX_NEW_INPUT_SYSTEM` define wired through `versionDefines`. No project-side action required either way. See `Documentation~/CameraSystem.md` for details.

#### Breaking changes

**Source-level API surface:**

- `IMapInformation` gained a new member: `TerrainInfo Terrain { get; }`. Any project that ships its own `IMapInformation` implementation must add this property — return an instance of `TerrainInfo` (defaults are fine for a non-terrain map). The built-in `MapInformation` already implements it.
- `TerrainData.ElevationValuesUpdated` is now declared `public event Action` (was a plain `public Action` field). External code can subscribe with `+=` / unsubscribe with `-=` as before, but **direct assignment (`= myCallback`) and direct invocation (`ElevationValuesUpdated()`) no longer compile**. Migrate to `+=` for subscription. The previous single-setter `SetElevationChangedCallback` (which silently wiped other subscribers) has been removed.
- `MapboxTileData.SetDisposeCallback` was replaced with `public AddDisposeCallback(Action)` / `RemoveDisposeCallback(Action)` (multicast). External code that previously assigned through reflection or an internal-visible-to dependency should switch to the add/remove pair; subscriptions are now safely shared across the up-to-16 render tiles that consume the same `TerrainData` instance.
- `TerrainData.IsDisposed` is a new `public bool` (read-only). Async readback paths check this before assigning into the data; external consumers may also want to consult it.
- `MapboxMapVisualizer` now implements two new interfaces:
  - `IMapVisualizer` (existing, no change).
  - `ITileLifecycleSource` (new, in `Mapbox.BaseModule.Map`): exposes `TileLoaded` / `TileUnloading` events, `MapInformation` property, and `ActiveTiles` as `IReadOnlyDictionary<UnwrappedTileId, UnityMapTile>` via explicit interface implementation. The public concrete `ActiveTiles` property still returns `Dictionary<,>` — no break for existing consumers.
- New opt-in marker interface `ITileLifecycleObserver` (in `Mapbox.BaseModule.Map`). Layer modules that need a visualizer reference for bookkeeping can implement it and the visualizer calls `AttachToMapVisualizer(ITileLifecycleSource source)` on them during `Initialize`. `TerrainLayerModule` implements it for its terrain-bounds tracker; other modules are unaffected.

**Custom imagery module:**

- `CustomTMSTile` constructor extended from 4 args → 6 args: added `bool invertY` and `bool isMapboxService`. The legacy 4-arg constructor (`urlFormat`, `tileId`, `tilesetId`, `useNonReadableTexture`) is preserved as an overload defaulting to `invertY: true, isMapboxService: false` — external subclasses keep compiling. New code should use the 6-arg form to control invert-Y behaviour and Mapbox-signed request routing explicitly.
- `CustomSource.CreateTile` now always returns a `CustomTMSTile` when `UrlFormat` is set, regardless of `InvertY`. Previously `InvertY=false` returned a plain `RasterTile`. If you relied on the plain-raster path with a non-empty `UrlFormat`, set `UrlFormat = ""` to keep that behavior — empty `UrlFormat` falls back to plain `RasterTile`.
- `CustomTerrainSource.CreateTile` got the same empty-`UrlFormat` fallback to plain `RasterTile` to match `CustomSource`.
- `CustomTerrainLayerModuleScript` now passes the user-configured `Settings.DataSettings` (cache size, retina flag, non-readable textures, data-zoom clamp) through to `CustomTerrainSource`. Previously a throwaway `ImageSourceSettings` was used and the Inspector values had no runtime effect. **This is a behavioral break for users who had `DataSettings` set to non-default values** — the values now actually apply.

**Terrain settings:**

- `TerrainInfo` removed two fields that were written-but-never-read: `IsEnabled` (was unused) and `Exaggeration` (no consumer). Remaining fields: `MinElevation`, `MaxElevation`. Plus new constants `DefaultMinElevation = 0f` and `DefaultMaxElevation = 5000f`.
- The previous `SetElevationChangedCallback` method on `TerrainData` was removed (replaced by the multicast event — see above).

**Camera API:**

- New `MapCameraBehaviour<T>` open generic abstract base class. Concrete subclasses (`SlippyMapCameraBehaviour`, `Moving3dCameraBehaviour`) ship with the SDK. Custom cameras must subclass with a closed-type generic parameter.
- New `MapInput.Teardown(IMapInformation)` virtual hook. Existing subclasses of `MapInput` must override and unsubscribe any events they registered in `Initialize` — without the override the default is a no-op (no break, but recommended).
- New `MapInput.TouchCountDecreasedThisFrame` protected property (read inside `UpdateCamera` to detect 2→1 finger pinch-end and re-seed drag origin).
- New camera state mutator: `MapInput.CurrentTwoFingerGesture` removed from earlier review iterations — moved entirely inside `TouchInputHandler` as a private enum. External code can't depend on it (was never publicly exposed in v3.0.x).

**Tile provider:**

- `UnityTileProvider.MaximumZoomLevel` is now silently clamped to **30** internally (was unbounded). Higher values would have overflowed `1 << (_maxZoom - zoom)` past 30. Practically unreachable in Mercator (pyramid maxes at 22-24), but any Inspector misconfig now produces a log warning instead of an overflow.
- `UnityTileProvider` now warns at construction if `MinimumZoomLevel > MaximumZoomLevel`.
- `TileNode.Set` internal-shape signature changed from `(id, worldCenter, scale, boundsHeight)` to `(id, worldCenter, scale, boundsBottom, boundsHeight)`. Internal API — only relevant if you subclassed `UnityTileProvider` and touched `TileNode` directly.

**Map visualizer:**

- `MapboxMapVisualizer.MaxMercatorZoom` is a new `public const int` (`= 22`). Replaces the magic literal in `DelveInto`. Available for external consumers that need to align with the SDK's pyramid cap.
- `MapboxMapVisualizer.MapInformation` is a new public property exposing the map info. Was previously a protected field only.

**Removed from package:**

- `TileProviderBenchmark.cs` was removed from the SDK package. It was an internal tool that shouldn't have been included. The file is `.gitignore`d so a local copy can be kept by developers who need it; users will not see it in the SDK.

#### New features

**Camera & input:**

- New camera system with two concrete behaviours:
  - `SlippyMapCameraBehaviour` — fixed camera, map moves underneath. Best for 2D-style maps, navigation overlays, AR/VR.
  - `Moving3dCameraBehaviour` — camera orbits a target point. Best for 3D city exploration, games, simulations.
  Both wrap a `MapCameraBehaviour<T>` generic base that handles the `IMapInformation` event subscription lifecycle.
- Touch input support: single-finger pan, two-finger pinch zoom, two-finger same-direction vertical tilt. Pinch/tilt are mutually exclusive via a per-frame gesture decision in `UpdateInputState`. Touch rotate is not implemented (mouse-only via right-click drag).
- Input System soft-dependency support. The SDK reads through an `IPointerInput` abstraction (legacy vs new system) selected at compile time by `MAPBOX_NEW_INPUT_SYSTEM`, with two `IInputHandler` implementations (mouse vs touch) selected by build platform — Android/iOS device builds use the touch handler; Editor and Standalone builds use the mouse handler. There is no runtime fallback.
- `Documentation~/CameraSystem.md` walks through the camera setup, gesture mapping, and the Input System opt-in.

**Tile provider / LOD:**

- LOD overhaul:
  - Forward-projected distance replaces euclidean distance.
  - Exponential split threshold replaces screen-fraction.
  - Acute-angle compensation (`DistToSplitScale`) reduces detail for tiles viewed at grazing angles, ported from the native SDK's `tile_cover.cpp`.
  - Result: tile counts at high pitch are 40–60% lower vs the prior algorithm with comparable visual fidelity near the camera.
- `FrustumBuffer` setting (world-space units; scales with `mapInformation.Scale`). Pre-loads tiles just outside the screen edges so panning doesn't pop in.
- Reusable scratch arrays (`_corners`, `_planes`, `_quadrants`, `_pool`) eliminate per-frame allocations in `GetTileCover` and `InternalUpdateCoroutine`.
- Sub-sea-level terrain handled correctly: `boundsBottom` derived from `MinElevation/scale`, so tiles below Y=0 (Death Valley, Dead Sea) no longer frustum-cull.

**Terrain rendering:**

- Burst-compiled decode of terrain-RGB into elevation float[] via `Sync/AsyncExtractElevationArray` Burst `IJob`s. Min/Max computed inline during decode (saves a second C# pass).
- Burst-compiled collider vertex-fill (`BuildColliderVerticesJob`) writing into a reusable `NativeArray<Vector3>`. Triangles regenerated only when `SampleCount` changes.
- Async PhysX collider bake (when `asyncBakeCollider = true`): `Physics.BakeMesh` runs on a worker thread; `Mesh.sharedMesh` assignment happens one frame later when the cooked-data cache is ready. Main-thread cost is near-zero.
- Shader-mode tiles share a single flat render mesh (`TerrainSharedFlat`) across the entire tile pool. Per-tile state (`_HeightTexture_ST`, `_TileScale`, etc.) goes via `MaterialPropertyBlock` instead of per-tile `Material` instances — saves `Material` clone allocations across pool churn.
- Bilinear elevation sampling in `TerrainData.QueryHeightData`, the collider vertex job, and the shader bilinear subgraph. Replaces the prior nearest-neighbor `(int)xx` truncation that produced visible stepping when render-tile mesh vertices sampled a denser sub-region than the data tile's effective resolution.
- `ElevatedTerrainShader.shadergraph` and `MapboxTerrainBilinearSampling.shadersubgraph` were restructured during the bilinear-sampling and normal-banding work (ridge-band fix uses a 2-texel offset, per Option A in the prior investigation). Lighting and displacement look correct on eye-test against the example scenes. If a project ships custom shader variants derived from these graphs, re-verify after upgrading.
- `TerrainInfo` exposes observed `MinElevation` / `MaxElevation` (in meters) on `IMapInformation.Terrain`. Updated automatically as terrain tiles load/unload. The tile-provider's AABB and `UnityMapTile.SetFallbackMeshBounds` consume this to size the frustum-cull volume.
- Conservative bounds fallback (`UnityMapTile.SetFallbackMeshBounds`, 10000m) applied immediately on terrain-data attach so shader-displaced geometry doesn't get culled before real Min/Max arrive — or permanently, when CPU extraction is disabled.
- Per-tile `MeshCollider` configurable via `TerrainColliderOptions` (in `ElevationLayerProperties`). Supports placing the collider on a dedicated child GameObject + Unity Layer.
- Bounds tracking is now an opt-in concern of `TerrainLayerModule` (via `ITileLifecycleObserver` → `TerrainBoundsTracker`). `MapboxMapVisualizer` itself is terrain-agnostic.

**Custom imagery:**

- New `CustomSourceSettings.InvertY` toggle: when `true`, Y is `(2^z - Y - 1)` instead of `Y` in URL substitution (TMS-style).
- New `CustomSourceSettings.IsMapboxService` toggle: when `true`, the tile request is signed with the Mapbox access token; when `false`, it's a plain HTTP request.

**Editor / Inspector:**

- New custom drawer for `[SimplificationFactor]` attribute: dropdown of distinct-grid presets with a live HelpBox describing the resulting vertex grid + per-tile quad count. Legacy non-preset values now show a "migrate to a preset" warning instead of being silently snapped. Multi-object editing and prefab-override-safe via `BeginProperty`/`EndProperty`.
- New custom drawer for `TerrainColliderOptions`: foldout with grayed-out sub-options when `addCollider` is disabled, further gated when `useDedicatedColliderLayer` is disabled.
- New custom drawer for `TerrainSideWallOptions`: same reveal pattern for skirt-wall settings.
- New custom drawer for `[GameObjectLayer]` attribute: Unity Layer dropdown for `layerId` fields.
- `TerrainSettingsInspectorHelper` warns when extraction is force-enabled by other settings (UseShaderTerrain off, or collider on) and renders the `ExtractCpuElevationData` checkbox as disabled+ticked. The previous OnGUI silent write to `boolValue` was removed — no more silently dirtying inspected assets. Scene-scan for vector / components modules is cached and invalidated on `EditorApplication.hierarchyChanged`.

**Other:**

- New demo scene demonstrating how to use POI information.
- New documentation files added.
- Reorganized Mapbox context menu for creating Mapbox ScriptableObjects.

#### Performance

- Burst-compiled terrain-RGB decode: significantly faster than the prior C# loop.
- Burst-compiled collider vertex fill with persistent `NativeArray` buffers.
- `ElevationArrayPool`: pools `float[]` buffers across cache evictions, capped at 32 per-size (`MaxPerSizeDepth`). Avoids ~256 KB / ~1 MB allocations per tile decode for 256² / 512² heightmaps. Returns past the cap drop to GC. In Editor, the pool clears on `[RuntimeInitializeOnLoadMethod(SubsystemRegistration)]` so stale buffers don't survive play-mode stop.
- `DelveInto` rewritten as iterative explicit-stack: eliminates `new bool[4]` + `new UnwrappedTileId[4]` allocations per cache miss (up to ~85 recursive calls × ~32 bytes before).
- `MapboxMapVisualizer.Load` reuses `unityMapTile.Children` lists across pool cycles instead of allocating a fresh list per cache miss.
- `TerrainBoundsTracker` recomputes Min/Max from `ActiveTiles` once per coroutine tick (deferred via a dirty flag) instead of on every tile event.
- `TerrainData._cachedWidth` caches `sqrt(ElevationValues.Length)` once on `SetElevationValues` — `QueryHeightData` no longer recomputes per query.
- `SimplificationFactorDrawer` caches preset+label arrays per `(Min, Max, VertexBase)` tuple and reuses one `GUIContent` for HelpBox height calculation (was allocating per repaint).
- `TerrainSettingsInspectorHelper` scene scan no longer runs `FindObjectsByType<MonoBehaviour>` on every OnGUI repaint.

#### Fixes

**Tile pipeline:**

- Fixed multicast dispose notification — shared `TerrainData` now notifies all consumers on eviction (previously the single-setter callback silently dropped 15 of 16 sharing tiles).
- Fixed `TerrainData.ElevationValuesUpdated` previously being a plain `public Action`, which let `SetElevationChangedCallback` silently wipe other subscribers. Now a proper multicast event.
- Fixed `TerrainSource.ExtractElevationValues` using the wrong overload of `ExtractHeightData` — the `Action<float[]>` overload + 1-arg `SetElevationValues` left `Min/MaxElevation` at 0 on every tile loaded through the initial async path. Now uses the `TerrainData` overload (which sets Min/Max inline).
- Fixed `TerrainSource.ExtractElevationValues` hanging forever when async readback completes after `TerrainData.Dispose` — the coroutine now subscribes to both `ElevationValuesUpdated` and the dispose multicast and completes on either signal.
- Fixed `Async/SyncExtractElevationArray` not guarding against `TerrainData.IsDisposed` — late-arriving GPU readback callbacks would assign into evicted data and leak the rented buffer. Now both check before renting and after the decode; the rented buffer is returned to the pool when dispose arrives in the window.
- Fixed `MapboxMapVisualizer.RecomputeTerrainBounds` (now in `TerrainBoundsTracker`) not being wired to `ElevationValuesUpdated` for shader-mode tiles — those tiles finish on texture-ready before CPU decode arrives, so the bounds never widened until extraction landed. Now a one-shot watcher per shared `TerrainData` triggers the recompute.
- Fixed terrain bounds remaining pinned at 0 when all loaded tiles were at/above sea level (previous accumulate-only `< terrain.MinElevation` never fired).
- Fixed terrain bounds monotonically inflating over a long session — `TerrainBoundsTracker` now recomputes from `ActiveTiles` on tile pool, so bounds tighten as the camera moves to lower-elevation regions.
- Fixed `UnityTileTerrainContainer.SetTerrainData(null)` silently leaking the prior `TerrainData`'s `ElevationValuesUpdated` and dispose-callback subscriptions. Detach now runs unconditionally before the null-return.
- Fixed `UnityTileTerrainContainer.OnDestroy` not removing the dispose callback (only the `ElevationValuesUpdated` handler) — a later eviction would fire the multicast into a destroyed tile.
- Fixed `UnityTileImageContainer.OnDestroy` being empty — same multicast-leak pattern as the terrain container.

**Terrain strategy / collider:**

- Fixed CPU-elevation render mesh leak via `MeshFilter.mesh` clone path. `CreateElevatedMesh` now reads `.sharedMesh` (the `.mesh` getter implicitly cloned the assigned shared mesh and orphaned the original). Per-tile mesh is now explicitly named `TerrainCpuMesh` so `UnityMapTile.OnDestroy` correctly disposes it.
- Fixed `_sharedFlatMesh` being `Clear()`'d on a shader→CPU transition when `sampleCount` or `_useTileSkirts` changed at runtime — would wipe geometry for every other shader-mode tile pointing at the shared mesh. Now allocates a fresh per-tile mesh first.
- Fixed `_sharedFlatMesh` leaking on a second `Initialize` call. `Initialize` now destroys the previous one before allocating.
- Fixed `MeshCollider.cookingOptions` only being set at component creation — when an existing collider's cooking options diverged from the bake's, PhysX silently re-cooked synchronously on assignment, defeating the async-bake path. Now assigned every call.
- Fixed `CompleteBakeAndAssign` not null-guarding the new mesh after `JobHandle.Complete()` — a parallel bake completion could `Destroy()` our mesh as its "previous" before we got to assign it.
- Fixed `_elevationNativeMirror` cache key being the `float[]` reference. `ElevationArrayPool` recycles buffers, so the same `float[]` could be re-rented for different data and produce a false cache hit. Now keyed by `TerrainData` reference.
- Fixed `RegisterCollider` deferred path adding a closure for every temp→final re-invocation, doubling async-bake work. Now de-dups by `(tile, data)`.
- Fixed deferred-rebuild closures rooting `tile` and `data` when `TerrainData.Dispose` arrived before `ElevationValuesUpdated` fired. Closures now self-detach on both signals.
- Fixed `_lastColliderBuild` (collider build cache) growing with dead `MeshCollider` keys — Unity's overloaded `==` doesn't reach `Dictionary<K,V>`'s reference equality. Opportunistic sweep + `Clear()` on `OnDestroy`.
- Fixed mesh-name string duplication: `TerrainColliderMeshName`, `SharedFlatMeshName`, `TerrainCpuMeshName` are now `public const` on `ElevatedTerrainStrategy`. The dedicated-collider-layer GameObject name renamed from `"TerrainCollider"` to `"TerrainColliderChild"` to disambiguate from the mesh name.

**Tile provider:**

- Fixed `UnityTileProvider.ShouldSplit` overriding the per-corner `offset.y` to `+cameraHeight` — correct in the native SDK's tile-coordinate system but flipped the sign of the vertical dot-product contribution in Unity's world units, triggering "always split" for downward-pitched cameras. Now uses ground-plane projection (`offset.y = 0f`).
- Fixed `UnityTileProvider.DistToSplitScale` dividing by `dz` with no zero guard — camera-at-ground (`dz=0`) produced NaN/Inf and pinned the tile to `MinimumZoomLevel`. Now short-circuits with `return 1f` when `dz < 0.0001f`.
- Fixed `UnityTileProvider.ShouldSplit` always-split for behind-camera corners recursing all the way to `_maxZoom` on tilted views with `FrustumBuffer` expansion. Now capped at `zoom < 18` (`AlwaysSplitSafeCap`).
- Fixed `UnityTileProvider.FrustumBuffer` tooltip not mentioning that the value is in world-space units that depend on `mapInformation.Scale`.

**Camera & input:**

- Fixed `MAPBOX_NEW_INPUT_SYSTEM` define gating being broken — the `versionDefines` entry that defined the symbol was uncommitted. Projects with "Active Input Handling = New" only would crash on every legacy `UnityEngine.Input.*` call.
- Fixed `MapboxExamples.asmdef` requiring `Unity.InputSystem` as a hard reference, breaking projects without the package. Now relies on the InputSystem asmdef's `autoReferenced: true` resolution; absent → no reference, no compile error.
- Fixed scroll-zoom delta magnitude differing ~10× between legacy and new input paths on Windows (legacy `Input.GetAxis` ≈ 0.1/notch, new `scroll.y` ≈ 120/notch). Normalized to ~0.1/notch on both.
- Fixed pan jump after a 2-finger pinch drops to 1 finger — the new "primary" touch could be at a different screen position than the lifted one. Now re-seeds drag origin on `TouchCountDecreasedThisFrame`.
- Fixed pinch and tilt detectors both firing on the same frame (each computed dominance independently and `GetPinchZoomDelta` updated `_previousPinchDistance` regardless of which won). Decision now made once per frame in `UpdateInputState`.
- Fixed pinch/tilt thresholds being in pixel units (DPI-dependent). Both magnitudes now normalized to `Screen.height`; pinch wins ties (no dropped frames on near-vertical pure-tilt drags).
- Fixed pinch state corruption across 2→3→2 finger transitions. EnhancedTouch reorders `activeTouches` on add/remove, so the next frame would compare distance against a different physical pair. Now tracks the touch ID pair and reseeds on change.
- Fixed `Moving3dCamera.Zoom()` NaN propagation through `preDistance`, `camDistanceToMouse`, and the lerp output. Top-down ortho / first-frame / `mapInformation.Scale=0` / camera curve evaluating to 0 all hit the guards now.
- Fixed `MapCameraBehaviour.Awake` NRE when no `MapBehaviourCore` exists in the scene. Logs an error and disables the component instead.
- Fixed `MapCameraBehaviour` never unsubscribing from `MapBehaviour.Initialized` — leak across scene reload.
- Fixed `MapCameraBehaviour` not tearing down `IMapInformation` event subscriptions registered by the camera core (`LatitudeLongitudeChanged`, `ViewChanged`, `SetView`) — the closures rooted the destroyed camera and its `Camera` Transform. New `MapInput.Teardown` virtual hook + `OnDestroy` wiring.
- Fixed `MapCameraBehaviour.OnMapInitialized` not tearing down a previous `MapInformation` if the event fires twice (defensive — not currently exercised by example scenes).
- Fixed `SlippyMapCamera.Initialize` dereferencing `FindObjectOfType<MapBehaviourCore>().transform` without a null check.
- Fixed `SlippyMapCamera.CenterLatitudeLongitude` being overwritten every frame even when nothing moved, with a `Mathf.Clamp((float)Latitude, …)` downcast losing precision at z18+. Now gated on `_output.HasChanged` and uses `System.Math.Clamp` on doubles.

**Map visualizer:**

- Fixed `MapboxMapVisualizer.RemoveUnnecessaryTiles` previously exempting `Filler` tiles from removal — caused orphaned-filler z-fighting. Now eligible for pooling; visual continuity preserved by the temp-tile filler protection pass in `Load()`.
- Fixed `MapboxMapVisualizer.OnDestroy` not detaching pending elevation-decode watchers — closures kept shared `TerrainData` rooted past the visualizer's lifetime. Tracker (now in `TerrainBoundsTracker`) handles this in its `Dispose`.
- Fixed `LoadingState.Filler` children of currently-loading temp tiles being pooled when they're still needed for visual continuity. Filler protection pass in `Load()` removes their ids from `_toRemove`.

**Custom imagery:**

- Fixed `CustomSource.CreateTile` throwing `ArgumentNullException` from `string.Format(null, …)` when `UrlFormat` was empty. Empty `UrlFormat` now falls back to plain `RasterTile`.
- Fixed `CustomTerrainSource.CreateTile` missing the same empty-`UrlFormat` fallback.
- Fixed `CustomTerrainLayerModuleScript` not passing the user's `DataSettings` through to the source — Inspector values had no runtime effect.
- Restored 4-arg `CustomTMSTile` constructor as a backward-compat overload (binary break for external subclasses otherwise).

**Editor / Inspector:**

- Fixed `SimplificationFactorDrawer` silently snapping legacy non-preset values to the nearest preset during `OnGUI` — dirtied every inspected asset on first paint.
- Fixed `SimplificationFactorDrawer` missing `BeginProperty`/`EndProperty` — broke prefab-override and multi-object editing.
- Fixed `SimplificationFactorDrawer.GetPropertyHeight` returning a different height than the rendered HelpBox content for legacy values — caused HelpBox to clip into the next Inspector row.
- Fixed `TerrainColliderOptionsDrawer.GetPropertyHeight` returning one extra `line + spacing` (cosmetic gap at the bottom of the foldout).
- Fixed `TerrainSettingsInspectorHelper` writing `child.boolValue = true` during `OnGUI` — silently dirtied every inspected asset. Now display-only (runtime evaluates `NeedsCpuElevation` directly).
- Fixed `TerrainSettingsInspectorHelper.SceneHasFeaturesNeedingCpuElevation` running `FindObjectsByType<MonoBehaviour>` on every Inspector repaint. Cached and invalidated on `EditorApplication.hierarchyChanged`.
- Fixed `Runtime/Mapbox/VectorModule/ComponentSystem/Editor/` missing an asmdef — editor scripts (`BuildingLayerVisualizerObjectEditor`, `ExtrusionOptionsDrawer`, `MapboxComponentsModuleScriptEditor`) were being folded into the runtime `MapboxComponentSystem` assembly, pulling `UnityEditor` references into player builds. New `MapboxComponentSystem.Editor.asmdef` (Editor-only, references the runtime asm + `MapboxBaseModule`) keeps them out.
- Fixed `MapboxMapBehaviour.CreateMapVisualizer` silently producing a magenta tile material in player builds when no `TileCreatorBehaviour` was assigned. The previous fallback used `Shader.Find` to construct a Material at runtime, but `Shader.Find` only resolves shaders referenced from a shipped Material or listed in *Project Settings → Graphics → Always Included Shaders* — so the shader was stripped from builds and `new Material(null)` produced magenta tiles. Replaced with a new `[SerializeField] protected Material _defaultTileMaterial` on `MapboxMapBehaviour`; the asset reference keeps the shader alive in builds. Demo scenes (`LocationExample.unity`, `ApiTest.unity`) ship with this field wired to `ElevatedTerrainMaterial.mat`. If both `_tileCreatorBehaviour` and `_defaultTileMaterial` are null, an `InvalidOperationException` is thrown with actionable instructions.
- Fixed `UnityTileProviderSettings.SubdivisionBias` defaulting to `0` in scenes saved before the field existed — Unity assigns `default(float)` for missing serialized fields and does NOT run the C# field initializer in that case, which collapsed `distToSplit` and forced the tile provider to render only the `MinimumZoomLevel` tile. New C# default is `0.6f`; `ISerializationCallbackReceiver.OnAfterDeserialize` upgrades legacy serialized `<= 0` values to `0.6f`.

**Other:**

- Fixed `TerrainData.QueryHeightData(Vector2)` and `(float, float)` overloads not guarding against null `ElevationValues` — shader-only mode (`ExtractCpuElevationData=false`) hit the NRE.
- Fixed `TerrainData.Dispose` not being idempotent — explicit `IsDisposed` bail at top.
- Fixed `MapboxTileData.SetDisposeCallback` previously silently overwriting other subscribers when one `TerrainData` was shared across 16 render tiles (the single-callback API). Replaced with multicast `AddDisposeCallback` / `RemoveDisposeCallback`.
- Fixed `1 << (_maxZoom - zoom)` shift overflow risk in `UnityTileProvider.ShouldSplit` when `_maxZoom` exceeded 30. Now clamped internally.
- Fixed vector layer range limits not being applied correctly during data processing.
- Fixed incorrect limit validation and bounds checking in the `LatitudeLongitude` struct.
- Fixed `MapInformation.Initialize` not resetting `Terrain.Min/MaxElevation` to defaults — previous run's values could bleed into a fresh session.
- Fixed Android IL2CPP player builds with **Medium** Managed Stripping Level silently dropping every SQLite tile cache insert with a `ConstraintForeignKey` error. Sqlite-net populates row values via reflection (`PropertyInfo.SetValue`), but IL2CPP stripped the auto-property accessor methods (`set_id`, `set_name`, etc.) on the cache's model classes — `SetValue` then became a silent no-op, every read returned `id = 0`, and downstream `tiles` inserts failed the foreign key on `tile_set`. The four SQLite model classes (`tiles`, `tilesets`, `offlineMaps`, `tile2offline`) and every one of their persisted properties now carry `[UnityEngine.Scripting.Preserve]`, and the same types are declared in `Plugins/link.xml` as a secondary safety net. Map rendering was unaffected (memory cache covered it) but the SQLite disk cache wasn't actually persisting under Medium. **Stripping support after the fix:** Low (and below) is the tested target; Medium has been briefly verified and seems to work; High is untested but expected to work given the same preservation. Additionally, the `Error inserting …` log line now includes SQLite's extended error code (`ConstraintUnique` / `ConstraintForeignKey` / `ConstraintNotNull` / etc.) so future cache-side failures are diagnosable from the first line.
- Fixed `MapboxMapVisualizer` writing `tile.transform.position` (world space) on every tile cover update, which silently reset any parent transformation on the `MapRoot` / `BaseTileRoot` hierarchy. Tiles are now positioned via `transform.localPosition`, so translating or rotating the map's root (e.g. anchoring it to an `ARAnchor`) is honored. For unchanged scenes — `MapRoot` at world origin with identity rotation — behavior is identical. Non-unit `MapRoot.localScale` is still unsupported; drive map scale through `MapInformation.Scale` instead.
- Fixed vector content (buildings, areas, roads from `VectorLayerVisualizer` and the ComponentSystem `AreaComponentVisualizer` / `RoadComponentVisualizer`) failing to follow the same MapRoot parent transform as the terrain tiles. Two distinct issues compounded each other. (1) Four call sites were writing world-space positions on the vector hierarchy: the per-layer `GameObjectOffset` y-offset in both component visualizers, the `Settings.Offset` application + `_layerRootObject.transform.position.x/z` reads inside `VectorLayerVisualizer.UpdateForView`, and `MapShifterCore.Update` writing `RuntimeGenerationRoot.position -= viewCenterPosition`. All converted to `localPosition`. (2) `Transform.SetParent(parent)` defaults to `worldPositionStays: true`, which makes Unity preserve world transform by writing inverse-parent-rotation into the child's `localRotation`. Auto-created roots (`BaseTileRoot`, `RuntimeGenerationRoot` in `UnityContext`, and every per-layer `_layerRootObject` + pooled entity GameObject in the vector visualizers) hit this path on initial parenting, ending up with a `localRotation` that counter-rotated the entire vector subtree against MapRoot. All five `SetParent` call sites now pass `worldPositionStays: false`, with explicit `localRotation = Quaternion.identity` on the visualizer roots as defensive belt-and-suspenders. Buildings, areas, and roads now rotate / translate with the rest of the map under AR/MR-style parent transforms.

#### Documentation
- New [`Documentation~/Migration-3.0-to-3.1.md`](Documentation~/Migration-3.0-to-3.1.md) walks through every source-level break, every behavioral change, and step-by-step verification when upgrading from v3.0.
- `README.md` rewritten and now serves as the single landing page (Overview.md merged in and removed). Adds a quick-facts header table, dependency listing, demo-scenes table, iOS / Android build sections with the v3.1.0 stripping notes, modules table, and a complete documentation index.
- `Documentation~/GettingStartedWithMapboxMapObject.md` gained a "Place a GameObject at a latitude / longitude" recipe — the most commonly asked indie use case (drop a marker / character / POI at specific coordinates), with the correct `localPosition` + `MapRoot` parenting pattern and the optional `TryGetElevation` step.

#### Tests
- Added editor-mode tests for `ElevationArrayPool` (rent / return / pool cap / null safety), `TerrainInfo` (defaults), and `MapInformation` (Initialize resets terrain bounds, lat/lon, idempotency). First test coverage for v3.1.0 surface — more coming in v3.2.


astojilj (approved):
Offline review done. LGTM

PR #1996, opened 2026-07-07
Fix Input System dependency so the SDK compiles out of the box on Unity 6+

brnkhy:
Summary:
On newer Unity projects (Unity 6 and up), this package failed to compile because of how we handled the optional Input System package. This PR makes the dependency explicit and reliable, so the SDK now builds correctly regardless of which input setup a project uses.

What changed:
- com.unity.inputsystem is now a proper package dependency, so Unity always installs it alongside the SDK.
- Input handling now follows the project's own Player Settings (Active Input Handling) instead of guessing based on whether a package happens to be installed.
- Updated docs/README to match the new behavior.

Impact on existing projects:
No action needed. Projects using the old (legacy) input setup will keep working exactly as before — this only fixes projects that were previously broken by the compile error.

Bumped to v3.1.1 with changelog notes for anyone upgrading.

@flasher297


flasher297:
Approving this.

Here's also Claude's review.
Not sure how useful it is =)

This is much better — actual diff content. Let me pull more of it since it got cut off before reaching any `.cs` files (the diff is huge — the PR touches 226 code files).Good — this pull was much richer than the changed-files view. I now have real diffs for `TerrainData.cs`, `ElevatedTerrainStrategy.cs`, `MapboxTileData.cs`, `MapboxMapVisualizer.cs` (partial), `MapboxMap.cs`, `MapInformation.cs`, `FileCache.cs`, `SqliteCache.cs`, `Tiles.cs`/`Tilesets.cs`, `ElevationArrayPool.cs`, `Response.cs`, `RectD.cs`, `MeshGenTaskWrapper.cs`, `package.json`-referenced changes in README, and more. The full diff is far larger than one fetch (it's several MB across 226 `.cs` files), so this covers the base module and terrain layer — the highest-risk area. Here's the review.

## Code review — findings grounded in the actual diff

**1. Undocumented breaking rename: `MapboxMap.MapUpdated()` → `ForceRedraw()`**
```csharp
- public void MapUpdated()
+ public void ForceRedraw()
```
This is a public API rename with no mention in the CHANGELOG's "Breaking changes" section, despite that section being unusually thorough about everything else (constructor arg changes, event type changes, etc.). Any external code calling `MapUpdated()` per-frame (the old doc-comment literally said "we currently use per-frame") will fail to compile, silently, with no migration note pointing at it. This should be added to the breaking-changes list or the migration guide.

**2. `TileToRelativeFilePath` cache-key format change silently orphans existing offline caches**
```csharp
- return string.Format("{0}/{1}{2}{3}.{4}", ..., tileId.X, tileId.Y, tileId.Z, ...);
+ return string.Format("{0}/{1}_{2}_{3}.{4}", ..., tileId.X, tileId.Y, tileId.Z, ...);
```
The underscore fix is correct — `"223"` was genuinely ambiguous between `(22,3)` and `(2,23)`. But this changes every cached file's path on disk. Existing installs upgrading to this version will have their entire file-cache silently become dead weight (old files never matched/cleaned, new files re-downloaded under new names). That's a real user-facing storage/bandwidth cost that isn't called out anywhere in the CHANGELOG or migration guide — worth a note plus maybe a one-time cache-migration or cleanup step.

Also, right next to it:
```csharp
private string TileToToFileInfoExpects(CanonicalTileId tileId, string tilesetId)
```
Typo'd name ("ToTo"), and it duplicates `TileToPathFileInfoExpects` byte-for-byte with no call site in this diff. Looks like leftover scratch code from the refactor — flag for removal.

**3. `Response()` constructor visibility change is a blunt fix**
```csharp
- private Response() { }
+ public Response() { }
```
This is presumably to satisfy a deserializer (Newtonsoft) that needs a parameterless constructor, but making it fully `public` removes all invariant protection the class had — any code anywhere can now construct a `Response` with `Request = null` and whatever other fields default to. A narrower fix (internal ctor + `[JsonConstructor]`, or a factory method) would have contained the blast radius. Worth checking whether anything downstream assumes a `Response` is never in a "manually constructed, half-initialized" state.

**4. Fragile self-referential closure pattern in `ElevatedTerrainStrategy.RegisterCollider`**
```csharp
(TerrainData data, UnityMapTile tile, Action rebuild) entry = (data, tile, null);
rebuild = () => { ...; _pendingRebuilds.Remove(entry); ... };
onDispose = () => { ...; _pendingRebuilds.Remove(entry); ... };
entry.rebuild = rebuild;
_pendingRebuilds.Add(entry);
```
This works because C# closures capture the *variable*, not a value snapshot — so `entry.rebuild` gets backfilled before either closure fires. But it's a genuinely fragile pattern: reordering these four lines (e.g., moving the `Add` above the `entry.rebuild = rebuild` assignment) would silently break de-dup and self-removal, and nothing would catch it at compile time. A small mutable class instead of a captured value-tuple would make the self-reference intent explicit and safer against future edits.

**5. `_elevationNativeMirror` cache key may go stale on same-instance re-decode**
```csharp
if (ReferenceEquals(terrainData, _elevationNativeMirror) && _elevationNative.IsCreated && _elevationNative.Length == source.Length)
{
    return; // skip re-copy
}
```
The commit message explains this was fixed to key by `TerrainData` reference instead of the recycled `float[]` reference — good instinct. But if the *same* `TerrainData` instance gets `SetElevationValues` called on it twice with an array of the *same length* (plausible: a retry, or a fetch superseding a stale in-flight one), this check would skip the mirror refresh and the Burst collider job would silently sample the old contents. A generation/version counter bumped in `SetElevationValues` and compared here would close this rather than relying on reference+length identity as a proxy for "content unchanged."

**6. Shared `_toRemove` `HashSet` reused across `Load()` and `InternalUpdateCoroutine()`**
Both methods `Clear()` and repopulate the same field. This is fine under Unity's single-threaded coroutine model as long as the two never interleave mid-execution, but it's an implicit invariant with no assertion or comment guarding it — a future change that adds a `yield` inside either method (easy to do accidentally while "just" awaiting a frame) could have one stomp the other's in-progress set. Worth a comment noting the non-reentrancy assumption, at minimum.

**7. Good, legitimately careful work:**
- `ElevationArrayPool`: proper lock discipline, capped depth, editor-only double-return detection via `[Conditional("UNITY_ASSERTIONS")]` — this is the right way to add a debug safety net without production cost.
- `TerrainData.ReadElevation`: bilinear sampling correctly clamps both source coordinates and the four sample indices; no off-by-one into out-of-bounds memory.
- `MapInformation.SetInformation`: added `!Mathf.Approximately(scale.Value, Scale)` before firing `OnWorldScaleChanged()` — small but correct fix to avoid spurious tile-reflow events from float noise.
- `MapboxTileData`/`TerrainData` dispose chain: multicast callback + `IsDisposed` idempotency guard is a clean, correct fix for the "16 tiles share one TerrainData, only one got notified" bug described in the CHANGELOG — the code matches the stated intent.
- `[Preserve]` attributes on the SQLite model classes: correctly targeted at the actual reflection-accessed members, plus the `link.xml` backstop — matches the stated IL2CPP stripping root cause exactly.

**8. Still open questions I can't verify from what I've pulled:**
- Whether `MapboxMapVisualizer.OnDestroy()` unsubscribes `_mapInformation.WorldScaleChanged` and the newly-added `LatitudeLongitudeChanged` handler — the diff cut off before I could see the tail of that method. Given how much of this PR is dedicated to fixing exactly this class of leak elsewhere, it'd be worth specifically confirming the visualizer doesn't have the same bug in its own teardown.
- The rest of `DelveInto`'s iterative rewrite and `UnityTileProvider.ShouldSplit`/`DistToSplitScale` — these are the files the commit messages flagged as containing the sign-flip and NaN-guard fixes, and I haven't pulled their actual diffs yet.

Want me to keep pulling — specifically `UnityTileProvider.cs` and the rest of `MapboxMapVisualizer.cs`/`DelveInto`, since those are the LOD/quadtree files with the highest bug-risk changes per the commit messages?

